PR #125, opened 2014-12-23
bug in README.md

sobi3ch:
Composer should require drupal/drupal**-**extension not drupal/drupal**.**extension``


sobi3ch:
Cool!

PR #129, opened 2015-01-26
Update requirements.rst

Elfiggo:
Help users to check if curl, mbstring and xml are installed by running php -m


jhedstrom:
Thanks!

PR #131, opened 2015-01-30
Use better technique to determine whether user is logged in

axe312ger:
Not all websites have the logout link on the front page, in my case, Chrome with Selenium was simply to fast,  admin-menu showed up after the check for the link happened.

Since there is no real 100% reliable method in D7, I've implemented two more methods to determine if a user is logged in. (D8 has the user information in the global Drupal Javascript object)

First test: Check for a "logged-in" class on the body element, most themes support this. (Mothership for example doesn't)
Second test: Visit "/user/login", if there is no login form, the user should be logged in.

I also had the Idea to simple check "/user/login" for a 403 status code, but this is not supported by Selenium because it is out of scope. (https://code.google.com/p/selenium/issues/detail?id=141)


src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -309,10 +309,26 @@ public function logout() {
    */
   public function loggedIn() {
     $session = $this->getSession();
+    $page = $session->getPage();
+
+    // Look for the logged-in class on the body tag.
+    // This should work with almost any theme.
+    $body = $page->find('css', 'body');
+    if ($body->hasClass('logged-in')) {
```
jhedstrom: We should make the `logged-in` class configurable, the same way the log out text is (see `getDrupalText()` usage below for example). (Same goes for the below `form#user-login`.)

jhedstrom:
I like this direction. This is a very-often-requested feature!

axe312ger:
Here we go. The two checks can now be altered in the selector section of the yml file. Please have a look :)

zuernBernhard:
I'd like to help on this. How can I run the travis-build locally (ubuntu linux) to debug the problem ?

jhedstrom:
I've never had much luck getting travis to run locally (there are various online blogs about it, but they get deprecated every 6 months or so).

However, you should be able to get the tests up and running locally by duplicating what is done in the `.travis.yml` file (particullary in the `before_script` and `script` sections).

Essentially:

* Get a Drupal 7 site up and running, and served by drush runserver
* Run the behat tests.

zuernBernhard:
Yes you are right. Hope I get the same problems like Travis does ;).

axe312ger:
I just wrapped the first check for the css class in a try {} statement. When using the Drupal api driver, there was no page loaded at this point and this caused the login process to break.

axe312ger:
I found no other way to check if the driver already loaded a page. Does anyone have a better approach?

zuernBernhard:
So we still need a logout-link or css to determine if there is a user logged in ?

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -335,12 +324,31 @@ public function logout() {
...
+    $page = $session->getPage();
+
+    // Look for a css selector to determine if a user is logged in.
+    // Default is the logged-in class on the body tag.
+    // Which should work with almost any theme.
+    try {
+      if ($page->has('css', $this->getDrupalSelector('logged_in_selector'))) {
+        return TRUE;
+      }
+    } catch (\Behat\Mink\Exception\DriverException $e) {
```
jhedstrom: Can we use a `use` statement instead of having the full namespace here?

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -283,7 +273,6 @@ public function userCreate($user) {
    */
   public function termCreate($term) {
     $this->dispatchHooks('BeforeTermCreateScope', $term);
-    $this->parseEntityFields('taxonomy_term', $term);
```
jhedstrom: This branch should merge in the latest master so these changes aren't reverted.

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -137,6 +136,21 @@ public function getDrupalText($name) {
...
   /**
+   * Returns a specific css selector.
+   *
+   * @param $name
+   *   string CSS selector name
+   */
+  public function getDrupalSelector($name) {
+    $text = $this->getDrupalParameter('selectors');
+    if (!isset($text[$name])) {
+      var_dump($text);
```
jhedstrom: `var_dump` should be removed.

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
axe312ger: Oops sorry, that was not my intention. I though I checked the commit first. Ill revert asap.

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
axe312ger: Of course we can :)

MKorostoff:
Pretty sure you'd want to look up the session ID and then check to see if there's a cookie named that, e.g.
 
```
$session_id = session_id();
$this->getSession->getCookie($session_id)
```

Rather than testing for something that may or may not be on the page.  I'm pretty sure that `session_id()` would be the wrong function there, just trying to show the general principle.

[Relevant docs](http://mink.behat.org/en/latest/guides/session.html#cookie-management)

jhedstrom:
That makes a ton of sense. Thanks @MKorostoff :+1:

recrit:
If anyone runs into this issue and needs a work around before a fix is committed, then the following assertions can be created in your FeatureContext that implement a customized loggedIn() check that can be changed as needed for your site.  The assertion code is a direct copy and paste from DrupalContext, and the text has the same beginning with an addition of " on this site".

```php

class FeatureContext extends RawDrupalContext implements SnippetAcceptingContext {
  /**
   * Check logged in status.
   *
   * Overrides RawDrupalContext::loggedIn().
   * @see https://github.com/jhedstrom/drupalextension/pull/131.
   */
  public function loggedIn() {
    $session = $this->getSession();
    $page = $session->getPage();

    // Body class check from pull/131.
    $body = $page->find('css', 'body');
    return $body->hasClass('logged-in');
  }

  /**
   * RawDrupalContext::assertAnonymousUser() with better logged in check.
   *
   * @Given I am an anonymous user on this site
   * @Given I am not logged in on this site
   */
  public function assertAnonymousUserOnThisSite() {
    // Verify the user is logged out.
    if ($this->loggedIn()) {
      $this->logout();
    }
  }

  /**
   * Creates and authenticates a user with the given role(s).
   *
   * RawDrupalContext::assertAuthenticatedByRole() with better logged in check.
   *
   * @Given I am logged in as a user with the :role role(s) on this site
   */
  public function assertAuthenticatedByRoleOnThisSite($role) {
    // Check if a user with this role is already logged in.
    if (!$this->loggedInWithRole($role)) {
      // Create user (and project)
      $user = (object) array(
        'name' => $this->getRandom()->name(8),
        'pass' => $this->getRandom()->name(16),
        'role' => $role,
      );
      $user->mail = "{$user->name}@example.com";

      $this->userCreate($user);

      $roles = explode(',', $role);
      $roles = array_map('trim', $roles);
      foreach ($roles as $role) {
        if (!in_array(strtolower($role), array('authenticated', 'authenticated user'))) {
          // Only add roles other than 'authenticated user'.
          $this->getDriver()->userAddRole($user, $role);
        }
      }

      // Login.
      $this->login();
    }
  }
}
```

axe312ger:
Just checked it out, it looks like the session cookie is only set if you are logged in. For some reason, I had in mind that this cookie is even set when you are logged out. I try to update this pull request asap :)

jhedstrom:
Setting this as a target for version `3.2.0`.

natgateway:
I am experiencing an issue that may be related to this one: I'm creating a custom step where I need the UID of the current user to check certain permissions, but it always returns as 0 (anonymous) even though my test generates a user with a particular role. However I'm using a background to generate the user for multiple scenarios.

      │     [uid] => 0
      │     [hostname] => 127.0.0.1
      │     [roles] => Array
      │         (
      │             [1] => anonymous user
      │         )
      │ 
      │     [cache] => 0

@recrit 
Thanks for posting that code snippet. I was wondering if that snippet would also work for test features where a user is created via background. I received a "Unable to access the request before visiting a page (Behat\Mink\Exception\DriverException)" error when the user is being created during the background stage.

My test.feature looks something like this:
        Background:
                Given I am logged in as a user with the "editorial" role on this site

        Scenario: When editorial navigates to node/add they should have access to the add/content page
                When I visit "/node/add"
                Then I should not see "access denied"
                Then I should see the heading "Add Content"
                Then <custom step here>

The test fails during the background stage. I'm wondering if that error is being triggered in the  loggedIn() function where $session->getPage(); is called.

My workaround for the time being is to run these tests against user that currently exist in the database although it's not ideal for repeat-ability.

I'm very new at behat/drupal testing so I very well may have missed something obvious, I'd appreciate any pointers on what to look for.

natgateway:
It turns out my problem was solved after getting some tips on this thread: https://github.com/jhedstrom/drupalextension/issues/173.

LionsAd:
I just ran into this with Drupal 8 and minimal profile, which does not have a logout link.

Interestingly it just started failing when using some Drupal Context tests.

I am not sure if it is possible or wanted to distinguish versions, but on Drupal 8 a logged in user can be easily identified by:

drupalSettings.user.uid being set to the user id of the user.

LionsAd:
https://github.com/jhedstrom/drupalextension/pull/237 is a replacement for this using the same code, but cleaned up, etc.

I propose to close this PR in favor of #237.

jhedstrom:
Agreed. Let's pursue this in #237.

PR #136, opened 2015-02-09
Add step definition to wait for Batch API to finish

jbickar:


jhedstrom:
@jbickar could you open this against the `master` branch, and perhaps move it into a new context, say `BatchContext`? This could also be a good addition to the examples part of the docs.

jbickar:
OK I'll try to take a look at doing so this week.

PR #137, opened 2015-02-10
Field handling implementation for API driver

ademarco:
The current pull request goes together with the correspondent Drupal Driver's pull request:

https://github.com/jhedstrom/DrupalDriver/pull/7

### Problem/motivation

Currently Drupal Behat extension provides limited support for attaching fields when creating nodes via the api driver (see https://github.com/jhedstrom/DrupalDriver/blob/master/src/Drupal/Driver/Cores/Drupal7.php#L299) and no support for fields for other kind of entities (e.g. users, taxonomy terms, etc).

At the moment the following test scenarios would fail:

```
  Scenario: Test node field handlers
    Given "page" content:
      | title      |
      | Page one   |
      | Page two   |
      | Page three |
    When I am viewing a "post" content:
      | title                | Post title                                               |
      | body                 | PLACEHOLDER BODY                                         |
      | field_post_reference | Page one, Page two                                       |
      | field_post_date      | 2015-02-08 17:45:00                                      |
      | field_post_dates     | 2015-02-10 17:45:00 - 2015-03-10 17:45:00                |
      | field_post_links     | Link 1 - http://example.com, Link 2 - http://example.com |
      | field_post_select    | One, Two                                                 |
    Then I should see "Page one"
    And I should see "Page two"
    And I should see "Sunday, February 8, 2015 - 18:45"
    And I should see "Tuesday, February 10, 2015 - 18:45 to Tuesday, March 10, 2015 - 18:45"
    And I should see the link "Link 1"
    And I should see the link "Link 2"
    And I should see "One"
    And I should see "Two"

  Scenario: Test user field handlers
    Given "tags" terms:
      | name      |
      | Tag one   |
      | Tag two   |
    And "page" content:
      | title      |
      | Page one   |
      | Page two   |
      | Page three |
    And users:
      | name     | mail         | field_tags       | field_post_reference |
      | John Doe | john@doe.com | Tag one, Tag two | Page one, Page two   |
    And I am logged in as a user with the "administrator" role
    When I visit "admin/people"
    Then I should see the link "John Doe"
    And I click "John Doe"
    Then I should see the link "Tag one"
    And I should see the link "Tag two"
    But I should not see the link "Tag three"
    And I should see "Page one"
    And I should see "Page two"
    But I should not see "Page three"

```

### Proposed solution

The proposed solution involves providing a flexible way of handling fields depending on their actual type and regardless of the entity they are attached to.

The solution involves modification to both Behat Drupal extension and Drupal Driver project. Separation of concerns between the two projects is carried out as following:

- Behat Drupal extension is responsible of defining a format to write scenarios containing multi-value, multiple fields (see example above for "field_post_links") in a human readable way. It is also responsible to parse such a format in order to provide the driver with a standardized array representing the field values.
- Drupal Driver receives such field values from Behat Drupal extension and, depending on the actual field type, it will expand the values in order to meet core or contrib API requirements.
- Drupal Driver attaches expanded fields to entities (i.e. nodes, users, taxonomy terms, etc.).

The current approach allows for the following step definitions:

#### Select fields
Being field_post_select a multiple select field with the following allowed values:

```
1|One
2|Two
```
we can set values using their labels instead of their keys:

```
| field_post_select    | One, Two |
```

#### Entity reference field
Being field_post_reference an entity reference field we can reference pages by just specifying their titles instead of their node IDs:

```
| field_post_reference | Page one, Page two |
```

At the moment only Drupal 7 version is provided since the current code, although working and tested, wants to get the discussion started on the issue in object.

The contributed code also provides a Drupal 7 feature that is required to be enabled on the test D7 in order to correctly run field handling test scenarios contained in ```field_handlers.scenario```.


ademarco:
Travis CI fails since composer.json still requires the original drupal drive and not the fork.

ademarco:
Ok build is actually working fine, not Behat fails on features/field_handlers.feature which is normal since it requires the bundled feature to be enabled on a clean Drupal 7 installation.

jhedstrom:
This is a really great direction! I love the cleanup in the Drupal Driver. Some of the fails (the changing of date format, for instance) look to be related to these changes.

ademarco:
True! I'll port the other scenarios too and re-run the test.

ademarco:
Actually I cannot reproduce locally the Travis CI fails, by running the same command I get:

```
$ ./vendor/behat/behat/bin/behat  -fprogress --profile=drupal7 --strict
......F-........................................F--................... 70
......................................................

--- Failed steps:

    Then I should be on "admin/structure/types/manage/article/fields" # features/api.feature:20
      Current page is "/drupal7/", but "/drupal7/admin/structure/types/manage/article/fields" expected. (Behat\Mink\Exception\ExpectationException)

    Given I am logged in as a user with the "administrator" role # features/api.feature:111
      Failed to log in as user 'aKSbZ6tC' with role 'administrator' (Exception)

26 scenarios (24 passed, 2 failed)
124 steps (119 passed, 2 failed, 3 skipped)
0m19.18s (47.73Mb)
```

I need to investigate a little more.

ademarco:
The commits above fix all tests for drupal7 profile, still drupal8 profile seems to be broken, even though locally it runs fine. 

Some side notes:

-It seems to not be necessary to revert the behat_test feature and clear the cache, I'll revert https://github.com/ademarco/drupalextension/commit/395b90c93eeeab31ca6a856b31fa48d28a50f041 commit
- Drupal 8 tests were failing only on Travis CI due to module_list() not being found (see https://travis-ci.org/jhedstrom/drupalextension/jobs/50233158#L576) so I've changed the way drupal 8 was downloaded in https://github.com/jhedstrom/drupalextension/commit/dffc8d30d06073b4c55b11a0cbbaf8fc33b9dc4a and now it still seems to fail, but for other reasons, see https://travis-ci.org/jhedstrom/drupalextension/jobs/50274284#L438

ademarco:
It seems that Drupal 8 build was failing since ```module_list()``` is not supported anymore by Drupal 8, I've fixed it here:

https://github.com/ademarco/DrupalDriver/commit/1cfbed11060282965795ea9eb29279a55bebb5dd

pfrenssen:
This looks pretty awesome. I haven't tested it yet but from just had a quick look at the code.

The test module can perhaps better be moved to a different location. We might also need to differentiate between test modules for different Drupal versions, this one only works for Drupal 7. Maybe put it in a new folder `fixtures/modules/Drupal7/`?

ademarco:
Sure, I'd do:

```
fixtures/Drupal7/modules/behat_test
fixtures/Drupal8/modules/behat_test
```

I'm working on the D8 version, I'll export configuration in a D8 module using https://www.drupal.org/project/config_devel so we can have similar TravisCI configuration for both Drupal cores.

ademarco:
Fix to the last fail is actually the following Drupal Drive fix: https://github.com/ademarco/DrupalDriver/commit/650bbc1d9f73ab4e5568b2ef87d968cd166c379f which didn't trigger another Travis CI build here. 

Would keeping ```composer.lock``` under VCS fix this issue (it is currently ignored)? If we do we would push the fact that DrupalDriver dependency has a newer commit, which will trigger a rebuild on this side too.

jhedstrom:
Once the PR on Drupal Driver is merged, the composer.json changes need to be backed out.

ademarco:
@jhedstrom @pfrenssen Basic Drupal 8 field handling is now available.

ademarco:
Ok, everything's green on the TravisCI side, we have a first version of both D8 and D7 field handling!

composer.json
```diff
@@ -11,13 +11,19 @@
        "email": "jhedstrom@gmail.com"
      }
   ],
+  "repositories": [
+    {
+      "type": "vcs",
+      "url": "https://github.com/ademarco/DrupalDriver"
+    }
+  ],
```
jhedstrom: I've merged the changes to the DrupalDriver into both master and 1.0, so this change can be backed out.

jhedstrom:
I'm going to merge this manually and make that change. Awesome work, thanks!

ademarco:
Thanks, that's great news!

ademarco:
I'll write some documentation over the weekend, what;s the best place to contribute that?

We should also refactor gists like https://gist.github.com/jhedstrom/5708233

jhedstrom:
Directly to [docs](https://github.com/jhedstrom/drupalextension/tree/master/doc) in the repo would be great! I'll take a look at updating that gist.

PR #138, opened 2015-02-11
Add documentation for environment specific settings

pfrenssen:
When running tests on different environments the environment specific settings need to be set in the `BEHAT_PARAMS` environment variable. This is not so straightforward so it would be nice to have documentation to help people with this.

This is my first foray into the wonders of the ReST text format and I have not yet found a good way of previewing this text so please forgive any formatting problems.

I have also rather shamelessly plugged my Drush extension that generates the environment variables automatically from the local environment.

### DIFF
--- a/doc/environment.rst
+++ b/doc/environment.rst
@@ -1,0 +1,42 @@
+Environment specific settings
+=============================
+
+Some of the settings in ``behat.yml`` are environment specific. For example the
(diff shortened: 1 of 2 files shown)


pfrenssen:
This is failing because of https://github.com/jhedstrom/DrupalDriver/pull/8

jhedstrom:
This looks good--I'll merge once the driver PR is merged.

PR #139, opened 2015-02-13
replaced "nodes" with "content" in example feature

stopfstedt:
see https://www.drupal.org/node/2403629


pfrenssen:
Good find!!

There is one more instance that currently mentions "node" but should be "content":

```
Scenario: Create and view a node with fields
    Given I am viewing an "Article" node:
```

stopfstedt:
thanks for checking! found another one that needed replacement.
please have a look again.

jhedstrom:
Thanks!

PR #140, opened 2015-02-13
Clean up DrupalSubContextInterface.

pfrenssen:
`DrupalSubContextInterface` could use a little love:

* The docblock for the constructor is not correct, it uses the wrong parameter type for `$context`.
* it states that the constructor returns a string but a constructor cannot return any value since it returns the object itself.
* We can also add some documentation to help people understand its purpose.


pfrenssen:
I also renamed the `$context` variable to `$drupal`. This better conveys what it contains, and this variable name is also used in the `FooFoo` example implementation .

PR #141, opened 2015-02-13
Provide a base class for subcontexts

pfrenssen:
It would be helpful if we provide a base class for subcontexts.

* Now all subcontexts have to implement the constructor but this is likely to be common for all subcontexts, causing code duplication.
* We can extend `RawDrupalContext` which would be a sane default for the most common case of subcontexts provided by contributed modules.
* If we have a base class we can also provide some helpful methods that can assist developers in building their subcontexts. As an example I have provided a method that allows the subcontext to get the currently logged in user from the DrupalContext class.

### DIFF
--- a/doc/_static/snippets/subcontext.inc
+++ b/doc/_static/snippets/subcontext.inc
@@ -1,16 +1,23 @@
 <?php
+
+/**
+ * Contains \FooFoo.
+ */
+
 use Behat\Behat\Tester\Exception\PendingException;
+use Drupal\DrupalExtension\Context\DrupalSubContextBase;
 use Drupal\DrupalExtension\Context\DrupalSubContextInterface;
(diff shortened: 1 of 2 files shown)


jhedstrom:
This looks great!

Could you re-push this now that the module error has been fixed in the drupal driver?

(I wish there was a way to re-queue travis builds.)

jhedstrom:
Oh, nevermind, Travis added a rebuild button!

openprivacy:
So where I used to extend RawDrupalContext I should now extend DrupalSubContextBase as a general rule?

jhedstrom:
If you're writing subcontexts you can extend this if you don't want to repeat all the boiler plate code it includes. For normal contexts, you can still extend `RawDrupalContext`.

PR #142, opened 2015-02-16
Do not register abstract subcontext classes.

pfrenssen:
This is a followup for [Provide a base class for subcontexts](https://github.com/jhedstrom/drupalextension/pull/141). When the list of subcontext classes is compiled we should ignore the abstract ones to avoid Behat from trying to register them.


jhedstrom:
Thanks!

PR #144, opened 2015-02-21
Add test coverage for user field handling in Drupal 8

ademarco:
The current pull request adds test coverage for user field handling in Drupal 8 by enforcing Behat Test module's user entity configuration in Drupal 8 installation, so that feature steps can be tested properly.


jhedstrom:
:+1: for more tests! Thanks!

PR #147, opened 2015-02-25
Update FeatureContext.php.inc

lhridley:
Correcting code snippet to reflect current FeatureContext.php file created by bin/behat --init


jhedstrom:
Thanks!

PR #150, opened 2015-03-02
Move building the extension list to the core driver

Berdir:
I'm still on 1.0, didn't check if this is already fixed in 3.0.

module_list() has been removed in 8.x, so the whole thing now depends on the core version, I moved it into the core drivers and added a new method for it.

### DIFF
--- a/src/Drupal/Driver/Cores/CoreInterface.php
+++ b/src/Drupal/Driver/Cores/CoreInterface.php
@@ -27,6 +27,14 @@ interface CoreInterface {
   public function bootstrap();
 
   /**
+   * Returns a list of extensions paths.
+   *
+   * @return string[]
+   *   List of paths to modules/themes/...
+   */
+  public function getExtensionPathList();
+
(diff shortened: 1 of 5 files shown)


jhedstrom:
This went in over on https://github.com/jhedstrom/DrupalDriver/pull/8 with a few changes after that was merged. I can backport your fix here if you change the method names to match what's already committed upstream.

Berdir:
Hm, but the fix there is different.

That just moved the module_list() call to the driver, I decided to move the complete logic, because drupal_get_filename() is also a drupal function and could potentially be different in different versions.

So, I can't just update the method names, because they're doing something different.

I see two versions, either going with the same approach as done there, or update both branches to use my approach, where the whole implementation is moved to the drivers.

jhedstrom:
That make sense. I'm merging this for now, and opened https://github.com/jhedstrom/DrupalDriver/issues/21 so this doesn't get lost upstream.

PR #155, opened 2015-03-19
Update requirements.rst

lesmanalim:
Several vendor includes require PHP 5.4, including Goutte and Guzzle. Also, Selenium Server is on 2.45.0.

### DIFF
--- a/doc/requirements.rst
+++ b/doc/requirements.rst
@@ -8,7 +8,7 @@ Meet the system requirements
 
     php --version
 
-   It must be higher than 5.3.5! Note: This means you cannot use the same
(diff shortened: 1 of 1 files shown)


jhedstrom:
My goal is to continue supporting 5.3 as long as possible, so what I'd ideally like in the requirements is what you have here, but also a big note about if php 5.3 is needed, then a custom `composer.lock` file is needed that will pin various dependent libraries at versions that support 5.3.

jhedstrom:
Closing this out for the reasons stated above. Thanks!

jhedstrom:
Note I think the next major version (`4.x`) will drop support for everything other than `5.6` and up, since even Drupal 6 can be run on `5.6`.

PR #158, opened 2015-03-23
Let's fix the Tests for d8

dawehner:
# Problem

you cannot longer override existing configuration from new installed modules

# Solution

* Move the override config files to "override_config"
* Override them using the existing BehatTestExtensionInstallStorage


jhedstrom:
Looks like a `cp -a` is needed here.

dawehner:
Failed test:
```
 Then I should see "Tag one" 
```
ha, that is bad.

jhedstrom:
Thanks!

Issue #162, opened 2015-04-15
Step not comparing attribute against :value argument


bt-eric:
For the step
`@Then I( should) see the :tag element with the :attribute attribute set to :value in the :region( region)`
the assertRegionElementAttribute function will return TRUE if it finds the attribute. It does not compare the attribute value against the :value argument. This is due to the break statement exiting the for loop before the code that checks the value is executed.

jhedstrom:
This has been merged. Thanks again!

PR #164, opened 2015-04-15
Issue #162 Step not comparing attribute against :value argument.

bt-eric:
This solves Issue #162 and https://www.drupal.org/node/2488352


jhedstrom:
Thanks!

Issue #135, opened 2015-02-04
Drupal user and content


Trogie:
Hello,

I had some feature files where I create content in certain scenarios with a user 'logged in with :role role'. After that scenario, I do check certain json calls. In a last scenario I did remove the content, again with 'logged in as a user with :role role' (another user offcourse than in the earlier scenario).

Everything worked ok but since a recent update (php composer.phar update) it looks like the content created with the first scenario gets deleted after that particular scenario...

Did something change recently in drupalextension? How to solve?

Trogie:
Hmmm my fault, had the drupal api commented out

Trogie:
Hmmm wasn't my fault but it looks like users get always deleted after a scenario with the 'user_cancel_delete' method.

I think the drupal driver should use the drupal settings from the site it is working on: my variable user_cancel_method is set to 'user_cancel_block' and so I don't expect the content to be removed with the user.

pfrenssen:
I'm still quite new to Behat but I believe this is intentional. After every scenario is tested all assets that were created during the scenario should be cleaned up so the website is back in the same state as before the test started.

This is why the 'user_cancel_delete' method is used: it ensures that all content created by the user is correctly removed.

There is a good reason for this: if any content or other assets stay behind then this might affect tests that run later. Imagine you have a test that creates 5 nodes with a certain keyword and then checks that the search results page contains 5 results for that keyword. This test would fail if any previous test would happen to create any node with that keyword.

If you need data created by a previous test scenario the best thing you can do is merge the two scenarios together. You can also use backgrounds to populate data to reuse in different test scenarios.

Trogie:
Off course, but why after every scenario? Why not after a Feature or complete test run? Why not make it configurable so the tester himself can decide?

pfrenssen:
Because every test scenario should be executable on its own. If your test scenarios are reusing data from previous test runs then they will fail if run in isolation. Also if you would make a change to test scenario 1 then suddenly test scenario 2 might fail because it relies on data from 1. It makes the test suite less reliable.

Take some time to go through the examples in the official Behat documentation: [Hooking into the test process](http://docs.behat.org/en/latest/guides/3.hooks.html). You'll see that these "@afterScenario" examples are mostly concerned with cleaning up data.

This is by the way not even a limitation of Behat, but it is common practice among all implementations of the Gherkin language. If you look up examples for Cucumber you'll also see that they clean up after each scenario.

Trogie:
[...]

PR #165, opened 2015-04-16
Adding a drupal config that allows you to set a configuration parameter that keeps the content created during a test from being deleted.

scottalan:
This would allow you to use tests to create test content.

Reference Issue: https://github.com/jhedstrom/drupalextension/issues/135

### DIFF
--- a/src/Drupal/DrupalExtension/Context/DrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/DrupalContext.php
@@ -178,6 +178,23 @@ class DrupalContext extends MinkContext implements DrupalAwareInterface, Transla
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jhedstrom:
The build is failing claiming it can't find the config `keep_content`, which is odd, since I can see it declared in the diff.

Trogie:
Very nice feature! Maybe add also an extra parameter so that content can also be deleted aftere the full feature, not after the scenario that created it?

scottalan:
@jhedstrom I did see the build failed after I made a PR but didn't look at the build until just now. Hmmm, yeah. It's there.

jhedstrom:
Closing this out as `1.x` is no longer supported, and due to inactivity.

PR #166, opened 2015-04-19
Add behat --init to the globalinstall documentation

pjonckiere:
I'm setting this up for the first time, and had to look around a bit to find out how to initialise behat since the globalinstall documentation doesn't specify that. We probably want to add that there.

The steps I followed were these: https://behat-drupal-extension.readthedocs.org/en/3.0/globalinstall.html


jhedstrom:
Thanks!

PR #168, opened 2015-05-06
"I am viewing my :type (content )with the title :title" generates notice on anonymous user

lpeabody:
### Changes
- Replaced `!$this->user->uid`with `!isset($this->user->uid)` which represses the notice and you will instead get the intended exception message.


jhedstrom:
Thanks!

PR #171, opened 2015-05-15
Update assertSelectRadioById

gturkalanov:
In order to fix "I select the radio button" step in Behat version 2+ we should replace "$this->getSession()->getSelectorsHandler()->xpathLiteral($label)" with "$label". Otherwise I receive the following error:
6384: The Behat\Mink\Selector\SelectorsHandler::xpathLiteral method is deprecated as of 1.7 and will be removed in 2.0. Use \Behat\Mink\Selector\Xpath\Escaper::escapeLiteral instead when building Xpath or pass the unescaped value when using the named selector. in vendor/behat/mink/src/Selector/SelectorsHandler.php line 131

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MinkContext.php
+++ b/src/Drupal/DrupalExtension/Context/MinkContext.php
@@ -508,7 +508,7 @@ class MinkContext extends MinkExtension implements TranslatableContext {
    */
   public function assertSelectRadioById($label, $id = '') {
     $element = $this->getSession()->getPage();
-    $radiobutton = $id ? $element->findById($id) : $element->find('named', array('radio', $this->getSession()->getSelectorsHandler()->xpathLiteral($label)));
+    $radiobutton = $id ? $element->findById($id) : $element->find('named', array('radio', $label));
     if ($radiobutton === NULL) {
       throw new \Exception(sprintf('The radio button with "%s" was not found on the page %s', $id ? $id : $label, $this->getSession()->getCurrentUrl()));
     }


jhedstrom:
Thanks!

pfrenssen:
I just updated BDE on a project and all my tests that use radio button labels are now failing. If I revert this commit they start working again.

jhedstrom:
I've reverted this and tagged new releases (`3.0.13` and `3.1.2`). Next time we'll need to add tests :)

pfrenssen:
Thanks! Let's fix it again and write a test indeed :)

Here's the issue that introduced the change: [#384: NamedSelector::translateToXPath error on 'content' selector](https://github.com/minkphp/Mink/issues/384). The actual commit that introduces the change is [e238faa7](https://github.com/minkphp/Mink/commit/e238faa7), which was included in release v1.6.0 of Mink. The maintainer of Mink, @stof, has added a comment in the issue explaining that this change introduces a BC break.

I have tested this and it's true: before the change you were supposed to pass an XPath escaped string to the named selectors, for example `'mystring'`. After the change you are expected to pass an _unescaped_ string such as `mystring` (without quotes).

This means that simply upgrading Mink from v1.5.0 to v1.6.0 will cause everybody's tests to break. Not only regarding radio buttons, but also any custom test code that uses named selectors. That's quite unfortunate. Luckily the work involved in updating the tests is very easy, people just need to pass the unescaped value, so all instances where a named selector is used:

```
$element->find('named', array('button', $this->getSession()->getSelectorsHandler()->xpathLiteral($label));
```

should become:

```
$element->find('named', array('button', $label));
```

How shall we proceed? We can stay on Mink 1.5 for the time being but that is not sustainable for long. We will need to bite the bullet at some point and force everybody to update their tests.

stof:
@pfrenssen no it won't break backward compatibility (otherwise it could not have been included in Mink 1.6 but only in 2.0). I found a way to accept both escaped and unescaped strings in Mink 1.6 (accepting escaped strings is deprecated though and will be removed in 2.0).

pfrenssen:
@stof, thanks for the info! I didn't realize this.

OK so then for the time being we just need to deal with the deprecated warning that is thrown on Mink 1.6+ and is making the tests fail. We can fix it on our end, we only have one call to `SelectorsHandler::xpathLiteral()` and when we update to Mink 1.6 we can remove this, but custom test code will still need to be updated.

I guess it would be acceptable to update to Mink 1.6 in the next minor release (so that would be 3.2.0) since the deprecated warning is very clear and easy to fix.

stof:
@pfrenssen we solved this issue in the upcoming Mink 1.7 though (to be released on Saturday hopefully): we now silence warnings as done in Symfony, so that they don't interfere with error handler (but error handlers following the Symfony convention for deprecation warnings will show them). I also plan to provide a way to have such an error handler in Behat to report deprecated usages.

Btw, I encourage you to upgrade to Mink 1.6 (your composer.json already allows it btw) as we did lots of driver consistency fixes in it (it was the main focus of the Mink 1.6 release)

stof:
On a side note, the change we are talking here is *not* part of Mink 1.6. It was done only after that (it is part of Mink 1.7.0, which has just been released)

PR #176, opened 2015-06-01
Use branch alias instead of branch

webflo:
I try to use another version of ``drupal/drupal-driver`` but the hardcoded branch name makes it unnecessary complicated.


jhedstrom:
Thanks!

PR #177, opened 2015-06-09
Allow users without roles

webflo:
I tried to create a user without any additional roles, but it did not work because we cast empty string to array.


jhedstrom:
The fails here are due to some sort of caching change in D8. Thanks!

PR #187, opened 2015-07-09
A build matrix for Travis CI

pfrenssen:
Let's try to get a nice build matrix up and running so we can test different Drupal versions on different PHP versions.


jhedstrom:
This is awesome! Thanks!

PR #190, opened 2015-07-17
Undesirable PHP Notice displayed when Exception thrown

nlisgo:
If we don't set a role on the user object then if the following Exception is thrown:

https://github.com/jhedstrom/drupalextension/blob/master/src/Drupal/DrupalExtension/Context/RawDrupalContext.php#L365

when we are checking if a user is logged in then it will error with:

```
Notice: Undefined property: stdClass::$role in vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/RawDrupalContext.php line 365
```

I've changed $rid to $role because it is the role name that is being stored on not the rid value.


jhedstrom:
Could you rebase this off the latest master? The tests are failing due to issues in drush master, and the latest code here pins drush to `8.*`.

nlisgo:
Will try and look at this tomorrow.

jhedstrom:
I've merged in the latest master in #434. Thanks for this fix!

Issue #193, opened 2015-07-23
Tests are failing since Drupal 8 updated Guzzle to 5.3.0


pfrenssen:
The Drupal 8 tests on the current master branch are failing because some interfaces have changed in the new Guzzle 5.3.0:

```
PHP Fatal error:  Declaration of GuzzleHttp\Client::send() must be compatible with GuzzleHttp\ClientInterface::send(GuzzleHttp\Message\RequestInterface $request) in /home/travis/build/pfrenssen/drupalextension/drupal/core/vendor/guzzlehttp/guzzle/src/Client.php on line 27
```

This is occurring because Behat needs Goutte, and Goutte needs Guzzle, so Guzzle is present in our autoloader. An older version of Guzzle is also present in Drupal 8 core, so when the D8 API functions are used they will invoke methods that are adhering to the older version, causing fatal errors to be thrown when an incompatible interface is encountered.

This is highlighting another problem: When Drupal 8 is bootstrapped in `DrupalDriver::getDrupalVersion()` the `autoload.php` of Drupal 8 is parsed, so we are actually loading two conflicting autoloaders, our own and Drupal's. This will cause conflicts when functions are redeclared, it is actually surprising that it worked up to now.

pfrenssen:
I was mistaken, it was not an upstream release of Guzzle, but actually the update of Guzzle in Drupal 8 that causes the problem. See [Issue #2493911: Update guzzle, goutte and mink-goutte-driver to the latest release](https://www.drupal.org/node/2493911).

Berdir:
Yes, I was confused a bit about 5.3.0

drupalextension should probably just update the requirements to get the same guzzle version? We had to update goutte and mink in core as well, so that should be possible.

pfrenssen:
I tried that but got problems with functions being redeclared when Drupal's autoloader is parsed.

```
Fatal error: Cannot redeclare GuzzleHttp\Psr7\str() (previously declared in drupal-extension/vendor/guzzlehttp/psr7/src/functions.php:17) in /drupal/core/vendor/guzzlehttp/psr7/src/functions.php on line 39
```

What would be the best way to solve this permanently? It seems like we cannot rely on Drupal's autoloader.

We at least need to keep all shared dependencies in sync, which will work for now but not when we start supporting Drupal 9 in the future.

pfrenssen:
[...]

PR #197, opened 2015-08-02
Provide a shorter alias for logging in as a user with a certain role.

pfrenssen:
This is a very simple feature request. One of the most common actions is to log in as a user with a certain role:

```
@Given I am logged in as a user with the :role role(s)
```

I would love to have this shorter version available as an alias. This read and writes a bit more fluently :)
```
@Given I am logged in as a/an :role
```


pfrenssen:
Tests are failing because of #193.

jhedstrom:
Thanks!

Issue #207, opened 2015-09-01
3.1.x requires PHP 5.4


iamEAP:
Looks like there were some bracket-style arrays introduced in the 3.1 release.  Should either swap those, or explicitly set a PHP requirement in composer.json.

```
Parse error: syntax error, unexpected '[' in /path/to/vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/RawDrupalContext.php on line 288
```

There may be others.

jhedstrom:
That wasn't intentional. We had to stop testing on 5.3 due to composer dependencies that were never sorted for 5.3. I opened #208 to see if we can do something similar to what we do for Drupal 8 with a pinned composer.json example.

jhedstrom:
PR is merged. I'm opening a follow-up to address the remaining tests that are failing (5.3 fails are currently configured to be ignored by Travis).

PR #208, opened 2015-09-01
Support PHP 5.3

jhedstrom:
- Addresses #207.


jhedstrom:
I'm going to merge this and cut `3.1.1`. I'll open a follow-up to fix the remaining test fails on php 5.3.

PR #213, opened 2015-09-16
Update globalinstall.rst re `FeatureContext` not found error/

gl2748:
When following the global install instructions to the letter I encountered an error at the last step.
The following error was displayed when I ran the behat -dl command:
    
    [Behat\Behat\Context\Exception\ContextNotFoundException]
    `FeatureContext` context class not found and can not be used.

Moving the FeatureContext.php file from /behat-tests/features/bootstrap to /behat-tests fixed the issue for me.

Hopefully this will be helpful to others.


jhedstrom:
Makes sense to me. Thanks!

PR #214, opened 2015-09-24
Little typo: use a right path

tanarurkerem:


jhedstrom:
The sample `composer.json` actually sets the bin directory to just `bin`, rather than `vendor/bin` (See  https://behat-drupal-extension.readthedocs.org/en/3.1/localinstall.html), so this is still in keeping with the current docs. I'm inclined to leave it at just `bin` for the time being.

tanarurkerem:
I saw it, but I concentrated only the README.md (See https://github.com/jhedstrom/drupalextension and try those steps), and I could not find how I change the bin directory when I use CLI. I worked harder, and I found it. :)

What do you think this pull request?

jhedstrom:
Ah, you're right. For the README, I think this change makes sense since it makes no reference to changing the bin directory. Thanks!

PR #215, opened 2015-09-28
Allow entity titles that include a comma

thewilkybarkid:
Currently entity titles are just split on a comma+space, which means that any that contain a comma can't be referenced.

This changes `explode()` to `str_getcsv()`, which should cater for this. It will, however, be a BC break for anything using titles that around surrounded with double quotes (they would now need to be escaped).

### DIFF
--- a/features/api.feature
+++ b/features/api.feature
@@ -146,16 +146,16 @@ Feature: DrupalContext
     | name      |
     | Tag one   |
     | Tag two   |
-    | Tag three |
+    | Tag,three |
     | Tag four  |
     And "article" content:
-    | title           | body             | promote | field_tags                  |
-    | Article by Joe  | PLACEHOLDER BODY |       1 | Tag one, Tag two, Tag three |
-    | Article by Mike | PLACEHOLDER BODY |       1 | Tag four                    |
+    | title           | body             | promote | field_tags                    |
+    | Article by Joe  | PLACEHOLDER BODY |       1 | Tag one, Tag two, "Tag,three" |
+    | Article by Mike | PLACEHOLDER BODY |       1 | Tag four                      |
     When I am on the homepage
     Then I should see the link "Tag one"
     And I should see the link "Tag two"
-    And I should see the link "Tag three"
+    And I should see the link "Tag,three"
     And I should see the link "Tag four"
 
   Scenario: Readable created dates
--- a/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
@@ -275,7 +275,7 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface {
 
   /**
    * Parse multi-value fields. Possible formats:
-   *    A, B, C
+   *    A, B, "C, D"
    *    A - B, C - D, E - F
    *
    * @param string $entity_type
@@ -313,7 +313,7 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface {
       if ($this->getDriver()->isField($entity_type, $field_name)) {
         // Split up multiple values in multi-value fields.
         $values = array();
-        foreach (explode(', ', $field_value) as $key => $value) {
+        foreach (str_getcsv($field_value) as $key => $value) {
           $columns = $value;
           // Split up field columns if the ' - ' separator is present.
           if (strstr($value, ' - ') !== FALSE) {


jhedstrom:
From the tests, this looks like it would need a corresponding change in the Drupal Drivers (https://github.com/jhedstrom/DrupalDriver)?

jonathanjfshaw:
I can't see why it would need a driver change?
But it should have a change to the test case: it should be "Tag, three" not "Tag,three" because it is comma space that is the delimiter without the quotes.

antonioG4:
I'm seeing this problem with not just the title field.  I'm trying to do the same with long text fields, or body field, and the same problem occurs.  I tried the proposed PR and it works for both titles and other text fields.

jonathanjfshaw:
The delimiter user is ", " not ",". Therefore the code needs to be `str_getcsv($field_value, ", ")`.
This should fix the test fails.

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -275,7 +275,7 @@ public function nodeCreate($node) {
 
   /**
    * Parse multi-value fields. Possible formats:
-   *    A, B, C
+   *    A, B, "C, D"
```
pfrenssen: This is very cryptic as documentation. Let's instead add a few lines of written text that clearly explain what is possible here.

features/api.feature
```diff
@@ -146,16 +146,16 @@ Feature: DrupalContext
...
     | Tag two   |
-    | Tag three |
+    | Tag,three |
     | Tag four  |
     And "article" content:
-    | title           | body             | promote | field_tags                  |
-    | Article by Joe  | PLACEHOLDER BODY |       1 | Tag one, Tag two, Tag three |
-    | Article by Mike | PLACEHOLDER BODY |       1 | Tag four                    |
+    | title           | body             | promote | field_tags                    |
+    | Article by Joe  | PLACEHOLDER BODY |       1 | Tag one, Tag two, "Tag,three" |
```
pfrenssen: Put a space behind the comma to make this a real example. It would also be clearer if we rename `Tag, three` to `Tag, containing a comma`.

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
jonathanjfshaw: How's this:
   * Parse a table row as a Drupal field, allowing for multi-value and compound fields. 
   * Compound field properties are identified using a ':' operator, either in the column
   * heading or in the cell. If multiple properties are present in a single cell, they must be
   * separated using ' - ', and values should not contain ':' or ' - '.
   * Multiple values in a single cell must be separated using ', ' and may be wrapped in
   * double quotes where necessary. 
   * Possible formats:
   * A
   * A, B, "C, D"
   * A - B
   * x: A - y: B
   * A - B, C - D, "E - F"
   * x: A - y: B,  x: C - y: D,  "x: E - y: F"
   * See field_handlers.feature for examples of usage.

thewilkybarkid:
I'm not actually using this anymore, but can make some tweaks to make it mergeable.

What about using `trim()` on all the values rather than saying the delimiter has the space?

jonathanjfshaw:
Your suggestion would treat both 'tag1,tag2' and 'tag1, tag2' as 2 values. This seems more friendly; the need for a space is something that has tripped me up repeatedly.

However, it also means that we can't create steps that deliberately have field values beginning or ending in a space, even using double quotes like 'tag1, " tag2"'. I can imagine wanting that in my steps in order to specify how such values should be treated.

Additionally, it is technically a change of behavior and a BC break, although a pretty trivial one.

We could devise a more complex regex to use instead of str_getcsv that explodes on either ',' or ', ', but still keeps the protective function of double quotes.

pfrenssen:
I agree that changing the space is a BC break, it's likely that there are tests out there in the wild right now that contain commas not followed by a space, and they would break if we change this.

I fully agree though that the space following the comma is quite arbitrary and can confuse people needlessly. I'd be happy to see it gone.

Now, we are actively working on the next 4.x release right now, so any BC breaks are fine, but then we'll not be able to backport this PR to 3.x.

We can also split off the space suffix to a separate PR that will be 4.x-only.

pfrenssen:
I just checked this again and we need the `trim()` because `str_getcsv()` keeps all spaces around the commas intact. This would otherwise break all existing tests that were forced to use the `, ` (comma + space) separator.

Let's add the `trim()` and just merge this in the 4.x branch only. I think that's the best way forward. Then 3.x has no BC breaks, and we are more friendly towards spacing the commas or not.

I have opened a new PR to continue this: #378.

PR #218, opened 2015-10-09
remove " in the step name

tikohatico:


jhedstrom:
Does this change work if `:rowText` has multiple words (eg, spaces)?

tikolakin:
Hi there, it  works in those ways
Then I should see username in the News row
Then I should see username in the "News Article" row
Then I should see "username" in the "News Article" row

when more than one word or spaces - quotes should be used.

jhedstrom:
Thanks!

PR #220, opened 2015-11-06
Use Random.name instead of Random.string for node and taxonomy content.

greg-1-anderson:
The characters present in Random.string make the Pantheon security sniffer suspicious when they pass through the [Behat Drush Endpoint](https://github.com/drush-ops/behat-drush-endpoint).

Perhaps it would be sufficient to simplify here?  This change shouldn't affect the robustness of the tests; if users specifically want to test certain characters in node creation, they could always explicitly use them in their test data.

Hope this is okay.


jhedstrom:
Makes sense to me. If folks are actually relying on the body field for certain tests, then they should probably be passing in explicit values anyway.

Thanks!

PR #222, opened 2015-11-10
Fix comparisons in tests; 

greg-1-anderson:
Former logic did not correctly set $? after test failures, so failing builds were mistakenly marked as passing.

We had:

test $v -eq 7 && behat || true

So, if $v was NOT 7, then the test returns false, and we need the `|| true` clause to prevent a spurious failure for the skipped test.  However, if $v IS 7, then the `|| true` clause runs regardless of the outcome of the behat test, so $? is always 0.

I changed this to:

test \! $v -eq 7 || behat

Now, if $v is NOT 7, then the test returns true, and processing stops, correctly skipping the test without marking it as an error.  If $v IS 7, then test returns false, and behat runs.  $? is set based on the outcome of the test.

Right now, the drupalextension tests are failing (my fault), but are being marked as successful due to this logic error.  So, if this PR is working correctly, then the test should correctly FAIL, and you should merge it.

Once this is right, I will submit a PR to fix the tests I broke.  I think it will need to go in the DrupalDriver, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,12 +35,12 @@ sudo: false
 install:
   - composer self-update
   # Use the example composer.json file for Drupal 8.
-  - test ${DRUPAL_VERSION} -eq 8 && cp doc/_static/composer.json.d8 ./composer.json || true
+  - test \! ${DRUPAL_VERSION} -eq 8 || cp doc/_static/composer.json.d8 ./composer.json
   - composer install
   # Drush version must vary depending on environment and version of core.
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Okay, test results are now matching expectations!

Also, I realized retroactively that there **is** such a thing as -ne, but "not the existing logic" was how I was approaching this, which is why I threw in the \!.  Could be converted to the other operator if desired.

greg-1-anderson:
After you merge this PR, you can make the tests green again by merging the above-referenced PR, https://github.com/jhedstrom/DrupalDriver/pull/68

jhedstrom:
Thanks! Merging the other one now.

PR #224, opened 2015-11-17
Provide steps to install and uninstall modules

pfrenssen:
I'm using DrupalDriver to test a distribution with a whole bunch of optional features. I would like to be able to install and uninstall them.

This pull request adds step definitions to install and uninstall modules. It relies on this sister PR in DrupalDriver: https://github.com/jhedstrom/DrupalDriver/pull/69

I have omitted the enabling and disabling of modules since this has been removed from Drupal 8 for good reasons. Since from an end user / BDD perspective there isn't much difference between installing and enabling a module I even made both terms available.

I've tested this only on Drupal 7 for now. Will test D6 and D8 tomorrow, and hopefully Drush too.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,13 +17,19 @@
     "behat/mink-selenium2-driver": "~1.1",
     "behat/behat": "~3.0,>=3.0.5",
     "behat/mink-extension": "~2.0",
-    "drupal/drupal-driver": "~1.0@dev"
+    "drupal/drupal-driver": "dev-install-uninstall-modules"
   },
   "require-dev": {
     "phpspec/phpspec": "~2.0",
     "phpunit/phpunit": "3.7.*",
     "behat/mink-zombie-driver": "^1.2"
   },
+  "repositories": [
+    {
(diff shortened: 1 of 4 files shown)


pfrenssen:
Many of the failures were due to the blackbox tests that needed to be updated since they relied on certain texts to be present on the homepage of drupal.org, but this has changed when Drupal 8 was released. This has been fixed in #230.

Will rebase against the latest master so we get a clearer view on what is failing.

pfrenssen:
OK it's looking better, the tests are now green on Drupal 7, still needs work for D6 and D8.

pwolanin:
@pfrenssen I've been looking for this kind of functionality, but I don't see how this PR works.  Seems like you need to implement the methods in the core drivers?

jhedstrom:
I'm inclined to make this a documentation issue (eg, how to do this in custom feature contexts, or elsewhere), since it is really low level stuff that is typically done in environment setup rather than in features/scenarios. Feel free to re-open for further discussion though.

Issue #225, opened 2015-11-18
Incompatible with symfony/yaml@3.0


mathewpeterson:
Symfony released version 3 which breaks backwards compatibility from 2.

As stated in the [symfony/yaml changelog](https://github.com/symfony/yaml/blob/master/CHANGELOG.md), it is required to quote `@` and ```

The yaml used in this project needs to be updated to reflect the breaking changes.

pfrenssen:
This has been fixed in PR #232 and can be closed. Thanks for the report!

PR #231, opened 2015-12-07
Provide an example that uses human readable field names

pfrenssen:
Recently support was added to Behat to do @transform on row-based tables: [PR #654: Add support for row table transformations](https://github.com/Behat/Behat/pull/654). This finally makes it possible to use human readable names for fields.

## Before

Up to now we were forced to use machine names which are not only ugly, but violate the BDD principle of using domain specific language:

```
When I am viewing a "post" content:
  | title                | Post title          |
  | body                 | Body text           |
  | field_post_reference | Page one, Page two  |
  | field_post_date      | 2015-02-08 17:45:00 |
```

## After

```
When I am viewing a "post" content:
  | title     | Post title          |
  | body      | Body text           |
  | reference | Page one, Page two  |
  | date      | 2015-02-08 17:45:00 |
```

Much better!

Note that this currently relies on code that has recently been committed to the master branch of Behat. We might want to wait on a new release of Behat before merging this in.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     "behat/mink": "~1.5",
     "behat/mink-goutte-driver": "~1.0",
     "behat/mink-selenium2-driver": "~1.1",
-    "behat/behat": "~3.0,>=3.0.5",
+    "behat/behat": "dev-master#2059cbe",
     "behat/mink-extension": "~2.0",
     "drupal/drupal-driver": "dev-master"
   },
--- a/doc/_static/composer.json.d8
+++ b/doc/_static/composer.json.d8
@@ -12,14 +12,15 @@
      }
   ],
   "require": {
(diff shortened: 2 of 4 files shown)


pfrenssen:
Tests on PHP5.5+ are failing because of #225.

pfrenssen:
Rebased on the latest master now that the fix for #225 is in.

pfrenssen:
It's failing on Drupal 8.

pfrenssen:
Going to work on this today.

pfrenssen:
Forgot to update the separate `composer.json` file which is used for the D8 tests. This should be green now.

I personally think it is best to await a new release of Behat before merging this in, so that we do not depend on a development version. I will maintain this branch until Behat gets a new release.

jhedstrom:
Thanks!

PR #232, opened 2015-12-07
Compatibility with Symfony 3.0

pfrenssen:
Pull request for #225.

On recent build jobs the following error occurs:

```
[Symfony\Component\DependencyInjection\Exception\InvalidArgumentException]                                                                                   
  The file "src/Drupal/DrupalExtension/ServiceContainer/config/services.yml" does not contain valid YAML.
                                                                                                                    
  [Symfony\Component\Yaml\Exception\ParseException]                                                                           
  The reserved indicator "@" cannot start a plain scalar; you need to quote the scalar at line 37 (near "- @drupal.random").
```

Starting in Symfony/Yaml 3.0 strings starting with an '@' need to be quoted.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     "behat/mink-selenium2-driver": "~1.1",
     "behat/behat": "~3.0,>=3.0.5",
     "behat/mink-extension": "~2.0",
-    "drupal/drupal-driver": "~1.0@dev"
+    "drupal/drupal-driver": "dev-master"
   },
   "require-dev": {
     "phpspec/phpspec": "~2.0",
--- a/src/Drupal/DrupalExtension/ServiceContainer/config/drivers/drupal.yml
+++ b/src/Drupal/DrupalExtension/ServiceContainer/config/drivers/drupal.yml
@@ -17,7 +17,7 @@ services:
     arguments:
       - %drupal.driver.drupal.drupal_root%
       - %mink.base_url%
-      - @drupal.driver.random
+      - "@drupal.driver.random"
     tags:
       - { name: drupal.driver, alias: drupal }
   drupal.driver.cores.6:
@@ -27,7 +27,7 @@ services:
     arguments:
       - %drupal.driver.drupal.drupal_root%
       - %mink.base_url%
-      - @drupal.driver.random
+      - "@drupal.driver.random"
   drupal.driver.cores.7:
     class: %drupal.driver.cores.7.class%
     tags:
@@ -35,7 +35,7 @@ services:
     arguments:
       - %drupal.driver.drupal.drupal_root%
       - %mink.base_url%
-      - @drupal.driver.random
+      - "@drupal.driver.random"
   drupal.driver.cores.8:
     class: %drupal.driver.cores.8.class%
     tags:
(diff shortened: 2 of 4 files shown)


pfrenssen:
This problem will not occur for people that use Behat 3.0.15 or lower since those versions still require `symfony/yaml ~2.1`. Support for Symfony 3 was added in [PR #801: Allow behat to work with Symfony3](https://github.com/Behat/Behat/pull/801), so until 3.0.16 is released it will only affect people that are using the current master branch.

Let's be ready in any case.

pfrenssen:
There were multiple affected Yaml files. There is also a strange error occurring which I cannot replicate locally:

```
[ReflectionException]
Class Drupal\Driver\Cores\Drupal6 does not exist
```

I've increased the debugging information on Travis, hope to get some pointers from that. It's strange that it even tries to load the `Drupal6` class for a Drupal 7 test.

pfrenssen:
For some reason Composer resolved DrupalDriver to version 1.0.8. In `composer.json` the version was set to `~1.0@dev` which I assume is intended to resolve to the latest development version on 1.x. On my system this works as expected. I've updated it to `dev-master` to see if this makes any difference.

pfrenssen:
Hah! This one's green but now it reverted back to `symfony/yaml` version 2.8.0! That explains why an ancient version of DrupalDriver was used: apparently something changed in the composer dependencies after 1.0.8 that makes it incompatible with Symfony 3.

pfrenssen:
This is probably it: [commit 84c1ff5c](https://github.com/jhedstrom/DrupalDriver/commit/84c1ff5c1ac9ede2554616c1c0ba87991eb6c46f). Will make a PR on the DrupalDriver side.

pfrenssen:
The last test run proves that it now works using Symfony 3.0 on PHP5.5+, but it didn't fall back to 2.8 on PHP5.4, hence the failure.

pfrenssen:
In any case the actual changes (the quoting of the yaml files) should be harmless on either version.

pfrenssen:
Woot! Persistence is the key to success!

Now that it's proven that it works, I'll clean up the branch.

jhedstrom:
I've merged https://github.com/jhedstrom/DrupalDriver/pull/72

pfrenssen:
Actually, let's wait for the sister pull request in DrupalDriver (https://github.com/jhedstrom/DrupalDriver/pull/72), then I can rebase this branch and get rid of all the exploratory commits easily.

pfrenssen:
Cool, thanks @jhedstrom ! I've rebased the branch and removed all unnecessary cruft. This does *not* prefer any Symfony version over another now, so the actual version used will be determined by the other dependencies that a project uses. At least it is now proven to be compatible with both.

jhedstrom:
From what I can tell, only certain Symfony 3.0 components are used here (finder for one). The YAML component is still at 2.8 for the travis builds. I think this is good for now though as mentioned above it has manually been confirmed to work with 3.0.

Thanks!

jhedstrom:
@pfrenssen are you able to run Behat, or the Drupal Drivers with Symfony 3? I am seeing increasing incompatibilities with Drupal 8 (see #244 test fails for instance).

pfrenssen:
I only have a Drupal 7 project that uses Symfony 3 dependencies. My D8 projects use Symfony 2.7 for Drupal 8.0.x and Symfony 2.8 for Drupal 8.1.x.

jhedstrom:
Ah, that makes sense. Thanks!

Issue #227, opened 2015-11-20
logged in as a user with permission not working?


morbus:
Scenario: The custom Ubercart order statuses are added.
    Given I am logged in as a user with the "administer order workflow" permission
    And I am on "admin/store/settings/orders/workflow"
    Then I should see the text "download"
    And I should see the text "shipping"

Results in:

Notice: Undefined property: stdClass::$role in vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/RawDrupalContext.php line 430

and the test fails.

morbus:
The role does appear to be created properly, and assigned to $this->user->roles (based on debugging in assertLoggedInWithPermissions), and that makes it to $this->login(). The actual Notice above is "just" a bug regarding $this->role being used instead of $this->user->roles(), but I'm still unable to figure out why the login (and the rest of the test) is failing.

morbus:
I'm able to confirm that the user, with this created role, is being logged in (by adding print $this->getSession()->getPage()->getContent(); after the $submit->click() inside $this->login()). But, once things get to $this->loggedIn(), it appears the page being returned into $element there, the user is logged out again...

morbus:
If I disable the if (!$this->loggedIn()) check, my test completes fine. Heh.

Berdir:
loggedIn() works by looking for a Logout link. Maybe you don't have that or you renamed it. There are some issues that discuss on how to improve that.

morbus:
@Berdir: It's not that, as I've many other logged-in-user tests working fine, even in the same Feature. I did think it might have to do with the page I'm requesting not having a Logout link on it, but the loggedIn() check moves to / first (which does have a Logout link).

morbus:
Smaller example of working/non-working definitions:

<pre>
  @runonlythis
  Scenario: The custom Ubercart order statuses are added.
    Given I am an anonymous user
    And I am on the homepage
    Then I should see the text "Featured titles"

  @runonlythis
  Scenario: The custom Ubercart order statuses are added.
    Given I am logged in as a user with the "content admin" role
    And I am on the homepage
    Then I should see the text "Featured titles"

  @runonlythis
  Scenario: The custom Ubercart order statuses are added.
    Given I am logged in as a user with the "access content" permission
      Notice: Undefined property: stdClass::$role in vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/RawDrupalContext.php line 430
    And I am on the homepage
[...]

PR #234, opened 2015-12-07
Improve verbosity of login failure.

fluxsauce:
This PR should address the usability and PHP error found in #227 

The check for a logged-in user is the same as Drupal's SimpleTest, which is the existence of a "log_out" link. I think it's fair to continue to look for that, but the error today doesn't describe the problem and has a PHP error if certain conditions are met.

Current:

> Failed to log in as user '%s' with role '%s'

Proposed solution:

> Unable to determine if logged in because 'log_out' link cannot be found for user '%s' with role '%s'

That way, developers know why the test is failing rather than something that implies that there's a problem with roles (there isn't).

Also, there's a PHP error if `$this->user->role` is missing, which was being experienced if you attempt to log in with permissions, not roles, which is why I'm checking for the existence of the property. Let me know if you think that's worth hunting down (I'm not familiar with it enough to know if it's working as designed).

### DIFF
--- a/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
@@ -427,7 +427,12 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface {
     $submit->click();
 
     if (!$this->loggedIn()) {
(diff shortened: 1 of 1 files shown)


fluxsauce:
Is it that because it's doing the permission-based login, `role` doesn't exist? `$this->user->roles` does exist.

jhedstrom:
Thanks!

fluxsauce:
Fantastic, thanks!

PR #236, opened 2015-12-08
[WIP] Extend test coverage for the D8 API and fix any uncovered problems

pfrenssen:
Currently the test coverage in `features/api.feature` mainly covers Drupal 7, but looking through the tests all this functionality should already be present in the D8 driver. Let's enable as many tests for Drupal 8 as are feasible.

### DIFF
--- a/features/api.feature
+++ b/features/api.feature
@@ -12,13 +12,21 @@ Feature: DrupalContext
     When I click "My account"
     Then I should see the text "Member for"
 
-  @drushTest @d7 @d8
+  @drushTest @d7
   Scenario: Target links within table rows
     Given I am logged in as a user with the "administrator" role
     When I am at "admin/structure/types"
     And I click "manage fields" in the "Article" row
     Then I should be on "admin/structure/types/manage/article/fields"
     And I should see text matching "Add new field"
+
+  @d8
+  Scenario: Target links within table rows
+    Given I am logged in as a user with the "administrator" role
+    When I am at "admin/structure/types"
+    And I click "Manage fields" in the "Article" row
+    Then I should be on "admin/structure/types/manage/article/fields"
+    And I should see text matching "Add field"
 
   @drushTest @d7 @d8
   Scenario: Find a heading in a region
--- a/features/api.feature
+++ b/features/api.feature
@@ -135,7 +135,7 @@ Feature: DrupalContext
     When I am viewing a "tags" term with the name "My tag"
     Then I should see the heading "My tag"
 
-  @d7 @d8
+  @d7
   Scenario: Create many terms
     Given "tags" terms:
     | name    |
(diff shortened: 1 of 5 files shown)


pfrenssen:
Ok so it looks like we have some work to do on Drupal 8 :)

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -160,6 +160,27 @@ public function getRandom() {
...
+    switch ($drupal_version) {
+      case 8:
+        foreach (array('changed', 'created', 'revision_timestamp') as $field) {
+          if (!empty($node->$field) && !is_numeric($node->$field)) {
+            $node->$field = strtotime($node->$field);
+          }
+        }
+      break;
+    }
+  }
```
pfrenssen: When saving nodes, the timestamps should be in UNIX time on Drupal 8, while Drupal 7 accepts a string. I've solved it here in a `@beforeNodeCreate` hook. Is this OK, or would it be better to handle this in DrupalDriver?

features/api.feature
```diff
@@ -72,14 +86,15 @@ Feature: DrupalContext
     And follow "First article with fields"
     Then I should see the text "PLACEHOLDER BODY"
 
+  @d7 @d8
   Scenario: Create and view a node with fields
-    Given I am viewing an "Article":
+    Given I am viewing an "article":
     | title | My article with fields! |
```
pfrenssen: This would be a nice followup: to allow the use of both machine names and human readable names of node types in this step definition.

features/api.feature
```diff
@@ -129,6 +168,20 @@ Feature: DrupalContext
...
+    When I go to "admin/structure/taxonomy/manage/tags/overview"
+    Then I should see "Tag one"
+    And I should see "Tag two"
+
+  @d7 @d8wip
+  # TODO: This doesn't work on Drupal 8 yet. For nodes the 'author' field is
+  # called 'uid' and only accepts numerical IDs.
   Scenario: Create nodes with specific authorship
     Given users:
     | name     | mail            | status |
```
pfrenssen: In order to create a node with a specific authors on Drupal 8 now we need to know the UID. We should make a followup for this. I guess this should be added to DrupalDriver: if the 'author' field is given when creating a node, convert this to the UID behind the scenes.

This is currently the only test that is not working for Drupal 8, all the others are running now.

features/field_handlers.feature
```diff
@@ -21,7 +21,9 @@ Feature: FieldHandlers
       | field_post_links     | Link 1 - http://example.com, Link 2 - http://example.com                         |
       | field_post_select    | One, Two                                                                         |
       | field_post_address   | country: BE - locality: Brussel - thoroughfare: Louisalaan 1 - postal_code: 1000 |
-    Then I should see "Page one"
+    Then I should see "Post title"
+    And I should see "PLACEHOLDER BODY"
+    And I should see "Page one"
     And I should see "Page two"
```
pfrenssen: Out of scope for this issue, but I noticed that two fields were not tested here, so I added them.

pfrenssen:
Oh dear we have a single failure on PHP 7 :(

features/api.feature
jhedstrom: That was possible at one point I thought.

pfrenssen:
OK I managed to install PHP 7 and can replicate the failure. This will probably have to wait until tomorrow though.

pfrenssen:
Got it! (I think :)

features/api.feature
pfrenssen: It seems to have worked fine in D7. I'll fix it in this PR and add test coverage for both the machine names and the human readable names so this won't regress.

jhedstrom:
Thanks!

PR #237, opened 2015-12-13
Use better technique to determine whether user is logged in #131

LionsAd:
Based heavily on https://github.com/jhedstrom/drupalextension/pull/131

This makes Drupal 7 work for themes without logout link on front-page and also makes Drupal 8 minimal profile, which uses stark work.

For most cases it can also avoid the visit to an additional page, which is great.

### Original report

Not all websites have the logout link on the front page, in my case, Chrome with Selenium was simply to fast, admin-menu showed up after the check for the link happened.

Since there is no real 100% reliable method in D7, I've implemented two more methods to determine if a user is logged in. (D8 has the user information in the global Drupal Javascript object)

First test: Check for a "logged-in" class on the body element, most themes support this. (Mothership for example doesn't)
Second test: Visit "/user/login", if there is no login form, the user should be logged in.

I also had the Idea to simple check "/user/login" for a 403 status code, but this is not supported by Selenium because it is out of scope. (https://code.google.com/p/selenium/issues/detail?id=141)

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MessageContext.php
+++ b/src/Drupal/DrupalExtension/Context/MessageContext.php
@@ -240,20 +240,6 @@ class MessageContext extends RawDrupalContext implements TranslatableContext {
   }
 
   /**
-   * Returns a specific css selector.
-   *
-   * @param $name
-   *   string CSS selector name
-   */
-  public function getDrupalSelector($name) {
-    $text = $this->getDrupalParameter('selectors');
-    if (!isset($text[$name])) {
-      throw new \Exception(sprintf('No such selector configured: %s', $name));
-    }
-    return $text[$name];
-  }
-
-  /**
    * Internal callback to check for a specific message in a given context.
    *
    * @param $message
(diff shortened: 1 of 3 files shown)


jhedstrom:
> Since there is no real 100% reliable method in D7, I've implemented two more methods to determine if a user is logged in. (D8 has the user information in the global Drupal Javascript object)

Any idea how much work it would be to add the D8-logic to the Drupal Drivers, and use that here (I assume it would not require making the redundant page load)?

jhedstrom:
Thanks for this!

dman-coders:
Thanks for this.
Took me an hour to trace and eventually deduce the same fix (admin-menu in use on D7 meant I was better to look for the logged-in class on the body)
 before I had enough background to trace the issue back to the issue queue to find myself here.

Now it seems I just have to use yesterdays git release. :-} Goody, I guess.

austinmcorso:
++ thank you for this

generalconsensus:
++ thank you for this! phantomjs was breaking because of the loggedIn() function

generalconsensus:
@jhedstrom @LionsAd Due to the inherent complexity of upgrading Behat to the latest dev release to be able to benefit from this functionality, would it be possible to patch earlier versions with this? Like say 1.0?

jhedstrom:
@generalconsensus I'd certainly review a PR for this fix against 1.x.

generalconsensus:
@jhedstrom Thanks! Here is the P.R. #247

PR #241, opened 2015-12-17
Turn on the Drush Driver tests for Drupal 8, and see how far we get.

greg-1-anderson:
Trying out the tests on Drupal 8, now that there have been some fixes made in the drush driver.


jhedstrom:
I think this is fixed now :)

PR #243, opened 2016-01-12
Adds drush output regex matching

acbramley:
Just a simple new step definition for matching drush output with regex, and a simple test to go with it.


jhedstrom:
Thanks!

PR #248, opened 2016-01-30
Add an assert to check if a button is not present on the page.

pfrenssen:
We have an `assertButton()` method to check if a form submit button is present on the page, but its counterpart `assertNotButton()` is missing.


jhedstrom:
Thanks!

Issue #258, opened 2016-03-07
Compatibility with big_pipe


Berdir:
I'm still on 1.x, but I'm having problems with the big_pipe module in 8.1.

Apparently behat/mink/goutte/guzzle doesn't follow the http-equiv refresh metatag that the module ads, and then it doesn't see messages or other "big-piped" blocks.

We should test if 3x still has that problem and possibly explicitly run the tests against 8.1 + big_pipe

Berdir:
As a possible Workaround, Wim suggested to force the bigpipe no-js cookie for all requests.

jhedstrom:
@Berdir so #259 is passing, but that doesn't actually mean this is fixed (we might just not be testing the right things--although there is a test for printed messages). I haven't been able to find any code in Guzzle, or Goutte, or Mink that indicates if `http-equiv=refresh` is actually used or respected. We might need a fix in the `RawDrupalContext`, or wherever else might make sense.

Berdir:
I have tons of tests that assert messages. For example, create a node and then check that the confirmation message is shown. Do you have any of those? They all fail for me.

jhedstrom:
@Berdir I added explicit tests over in #259 for messages, and they are still passing.

jhedstrom:
#259 has been merged, not sure what else can be done here without a way to make the tests fail with big pipe enabled.

zaporylie:
> BigPipe currently only is enabled for requests with a session.
see https://www.drupal.org/node/2657684

#259 only test messages for anonymous user.

mglaman:
I know this is stale, but I hit this when needing to test local tasks block.

```
<div data-big-pipe-placeholder-id="callback=Drupal%5Cblock%5CBlockViewBuilder%3A%3AlazyBuilder&amp;args[0]=bartik_local_tasks&amp;args[1]=full&amp;args[2]&amp;token=7daf9ccc"></div><div id="block-bartik-content" class="block block-system block-system-main-block">
```

But tests fail. On `"drupal/drupal-extension": "^3.2"` 3.2.2

mglaman:
The PR #259 with the `BigPipeContext` fixed my problems.

leymannx:
You mean #325? It has the `BigPipeContext`, which also fixed it for me.

Strange enough I had two similar scenarios, one with a publisher user role, one with an editor user role, both log in, access a random node and should click the "Edit" button. For the publisher this worked, for the editor it doesn't.

```feature
[...]

PR #259, opened 2016-03-07
Test Drupal 8 with big_pipe enabled

jhedstrom:
- Addresses #258

### DIFF
--- a/features/messages.feature
+++ b/features/messages.feature
@@ -1,0 +1,19 @@
+@api @d8
+Feature: Ensure that messages are working properly on local installs
+  In order to be sure that Drupal 8 with Big Pipe enabled can be tested
+  Messages are tested against a local installation
+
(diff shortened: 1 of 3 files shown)


features/messages.feature
```diff
@@ -0,0 +1,19 @@
+@api @d8 @arb
```
pfrenssen: @arb?

jhedstrom: Lol, I was using that for local testing as the behat command seems broken when trying to run line numbers of features combined with `--profile` :)

arb: Yes? :smile:

jhedstrom: Ha!

pfrenssen: Hahaha! Hello there!

arb: This LGTM. I say :shipit:

PR #263, opened 2016-03-25
Do not assume nodes of all types have 'body' fields

bartfeenstra:
All nodes have titles and types, but not all of them have bodies. This PR removes the assumption that this is so.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/DrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/DrupalContext.php
@@ -222,7 +222,6 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext {
     $node = (object) array(
       'title' => $title,
       'type' => $type,
-      'body' => $this->getRandom()->name(255),
     );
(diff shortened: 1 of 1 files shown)


jhedstrom:
This makes sense. The failing tests that rely on there being a body field will need to be updated I think.

bartfeenstra:
I stopped using these methods since they're hard to work with, and switched to creating entities using the Entity API directly. Therefore I cannot spend time on updating the tests for this PRs.

jonathanjfshaw:
I don't think there are any failing tests here, just outdated travis build failures. Should be green if tests are rerun.

Another reason why this patch is good is that the previous approach doesn't set the text format for the body field, so you end up with a blank format value which no node created in normal site use would have. This bit me today.

pfrenssen:
Good point, rebased this on the latest master branch and trying a new test run in #365

I started a new PR since @bartfeenstra has indicated he doesn't have the time to follow up on this. Would still like to have this in though.

Let's continue in #365,

PR #264, opened 2016-03-31
changed version of mink, behat, goutte

tikohatico:


jhedstrom:
Thanks for this. However, as you can see from the failing tests, these updates prove problematic to implement at the library level. Individual projects can use higher versions, but if I add these to the library, then anybody still testing a D6 site on php 54 (etc) would lose the ability to use this.

That being said, I have considered dropping php 53 and 54 support soon--but probably not until 4.0.

tikohatico:
Yep, the purpose of that was getting tests reports and decide what version to use.
Let's wait new release 
Thanks you

PR #268, opened 2016-04-19
Step to create queue items

greggles:
This PR allows for steps like...

```
  @api @queue
  Scenario: Items in the queue longer than 24 hours are detected
    Given there is an item in the system queue:
      | name       | somequeue    |
      | created    | 1460846596   |
    When I run drush "-v find-zombie-queue-items" expecting errors
    Then drush had error "The queue somequeue has 1 items, the oldest of which is Sat,"
    And drush had error "04/16/2016 - 15:43"
```

The last 3 steps are custom and not included, but I guess they are self-explanatory.

The default `data` is meaningless. I'm open to something else to put in for default.


jhedstrom:
The `FeatureContext` that ships with this extension is actually only intended for testing the extension itself. To add this as a re-usable step definition, it should be added as a new context (or potentially added to the existing `\Drupal\DrupalExtension\Context\BatchContext`). Then, the tests could be updated to run that new step definition.

greggles:
Thanks for the feedback and pointers. I've moved the step into `BatchContext` and also clarified how to use it and removed an unnecessary reference to rpid (which means something to me, but probably not most drupalextension users.

jhedstrom:
It's been some time :) This is looking good but needs to be rebased to get tests passing. It would also be good to note it in the `docs/contexts.rst` file :)

greggles:
It has been some time :)

Thanks for your feedback - now addressed along with some other bits of cleanup.

greggles:
No conflicts and checks have passed.

I wish there were some way for me to remove the "needs work" label :)

jhedstrom:
Awesome. Thanks!

PR #270, opened 2016-04-19
Update api.feature

d34dman:
Completing the scenario "Create an article with multiple term references"


jhedstrom:
Is this just for better documentation? Also, could you re-open against the `master` branch? Thanks!

PR #273, opened 2016-04-21
Provides "not in row" step definition and tests

### LINKED ISSUE #272: Provide step definition for "I should not see X in the Y row"
For testing things like access control on links on admin pages.

For example: We have custom access on editing nodes based on workbench_moderation states, we want a behat test to say that an author shouldn't see the edit link in a row of a particular node, which doesn't seem possible at the moment.

acbramley:
Fixes #272

### DIFF
--- a/features/api.feature
+++ b/features/api.feature
@@ -106,20 +106,24 @@ Feature: DrupalContext
   @d7
   Scenario: Create users with roles
     Given users:
-    | name     | mail            | roles         |
-    | Joe User | joe@example.com | administrator |
+    | name      | mail             | roles         |
+    | Joe User  | joe@example.com  | administrator |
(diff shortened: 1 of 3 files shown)


acbramley:
Seems to be some drush issues with the PHP 5.4 and PHP 5.5 (on Drupal 6) tests. Will wait for feedback from @jhedstrom

aronbeal:
@acbramley A Drush fix was [recently merged](https://github.com/jhedstrom/drupalextension/pull/277).  I was experiencing the same error pattern with my commit.  Re-basing off latest master fixed the issue.

acbramley:
@aronbeal thanks for that, giving it a shot now.

jhedstrom:
Thanks!

PR #275, opened 2016-04-26
For compatability with drupal 8.1.x

acbramley:
Not sure if there's a reason to pin these versions, mainly opening this up for the conversation and to see what breaks.

See https://github.com/drupal-composer/drupal-core/blob/8.1.x/composer.json#L10

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,8 +18,8 @@
     "behat/behat": "~3.1.0-rc2",
     "behat/mink-extension": "~2.0",
     "drupal/drupal-driver": "dev-master",
-    "symfony/dependency-injection": "~2.7",
-    "symfony/event-dispatcher": "~2.7"
+    "symfony/dependency-injection": "2.8.*",
+    "symfony/event-dispatcher": "2.8.*"
   },
   "require-dev": {
     "phpspec/phpspec": "~2.0",


composer.json
```diff
@@ -18,8 +18,8 @@
     "behat/behat": "~3.1.0-rc2",
     "behat/mink-extension": "~2.0",
     "drupal/drupal-driver": "dev-master",
-    "symfony/dependency-injection": "~2.7",
-    "symfony/event-dispatcher": "~2.7"
+    "symfony/dependency-injection": "2.8.*",
+    "symfony/event-dispatcher": "2.8.*"
```
jhedstrom: This would break compatibility with `8.0.*` (and future compatibility when core upgrades symfony). By using the `~`, individual users of this library are able to pin exact versions if needed.

acbramley: @jhedstrom good point. So in my own project's composer file I can just override the symfony requirement since the drupal-extension is using ~? I thought I had tried this but composer kept complaining.

jhedstrom: @acbramley yeah, as long as the versions you pin are compatible with Drupal core's versions that should work.

acbramley: @jhedstrom just updated that in our project and it worked a treat, thanks!

PR #280, opened 2016-05-09
Update RawDrupalContext.php

aronbeal:
Hey,

I completely missed up until this point that you were using the ':' character as an indicator of a multi-value field.  I'm offering the proposed enhancement to parseEntityFields, which hopefully will detect (in >= D7) when a multi-value field has an incorrect label (missing ':' character)

Thoughts?


aronbeal:
Cancel this for now, sorry.  I'm reviewing to make sure I understand the code.

PR #283, opened 2016-05-20
Add a quit handler to breakpoint

aronbeal:
Currently, the existing breakpoint function in DrupalContext waits on a carriage return key to continue.  If you're using this breakpoint to inspect a particular test, and you wish to exit after you have determined the test has failed, the existing method offers no option for it.  You could use Ctrl+C (on POSIX), but in my experience, db cleanup operations are improperly handled after a SIGINT.

This change side-steps the problem by throwing an exception after a 'q' entry by the user.  This in turn results in a test exit with a proper cleanup.


jhedstrom:
This is a great addition, thanks!

Issue #288, opened 2016-06-29
Alternate scenario outline steps are failing with a logged in user


Chapabu:
So, I have a scenario outline that checks for a selector on an admin page as follows.

``` gherkin
  Scenario Outline:
    Given I am logged in as a user with the "Content contributor" role
    And I am creating a "landing-page" node
    Then the "<Paragraph type>" paragraph type should be available on the "paragraph_reference" field.

    Examples:
      | Paragraph type      |
      | Heading             |
      | Highlights          |
      | Image               |
      | Quote               |
      | Video               |
```

This works fine on `v3.1.5`, but fails when updating to `v3.2.1`. Oddly enough though, it only fails on alternate steps. So in the example above, `Heading`, `Image`, and `Video` pass, whereas `Highlights`, and `Quote` fail.

I did a little bit of debugging, and it appears that the user is not logged in for the failing steps.

I'm not 100% sure where the problem lies, but I was hoping someone could either point me in the right direction as to which change in the release is causing this. I had a quick scan through, but figured I'd add an issue in case it's something someone has run into before.

djdevin:
I just got it too, and it's as you described - only happens on scenario outlines. Haven't debugged it yet.

jhedstrom:
Is it that last step that is failing? Does running with `--verbose` provide any additional information?

stmh:
I am experiencing a similar problem with drupal 8 and interacting with user-entities in scenario outlines.

I get the error because on @afterScenario all automatically created users gets deleted, but you are still logged in with the now non-existing user.

See pull-request #289

PR #289, opened 2016-06-30
Fix scenario outline #288

stmh:
When using scenario outlines and a logged in user, the user gets invalid on the second run of the oultine, as @afterScenario deletes all created users. The proposed fix invalidates the user-property and logs the user out.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
@@ -231,6 +231,10 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface {
       }
       $this->getDriver()->processBatch();
       $this->users = array();
+      $this->user = FALSE;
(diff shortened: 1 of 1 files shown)


src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -84,6 +84,7 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface {
 
   /**
    * Keep track of any languages that are created so they can easily be removed.
+   * Keep track of any languages that are created so they can easily be removed.
```
Chapabu: Did you mean to duplicate this line?

jhedstrom:
This fix makes sense. Can you remove the duplicated docs line mentioned above? Then I think this is good to go.

stmh:
Sorry for the duplicated line. Did remove it...

jhedstrom:
Thanks!

PR #290, opened 2016-07-03
Add MessageContext

wheatpenny:
In walking through the steps on the Blackbox Driver page, I realized that MessageContext needed to be added to the behat.yml file in order for the "Message selectors" section to function. I added the line to the snippet and updated the line references in the documentation page.


jhedstrom:
Thanks! This needs to be against `master` however, and can be backported to the latest `3.x` version.

PR #297, opened 2016-08-04
[TASK] French translation

mickaelperrin:
Hi,

This is a reopen of #291 but with a branch and not a simple commit.

I feel this translation will need update, I found that some translations where conflicting with mink.

I will update this later.


jhedstrom:
This was done in #431. Thanks for this though!

Issue #194, opened 2015-07-27
Given (a/an) :entity entity/entities/item/items


jonathanjfshaw:
Following on from: https://www.drupal.org/node/2537618, planning for adding support beyond nodes to the more basic content entities.

JHedstrom commented there:
"In the Drupal Drivers, there would be new methods added. For the changes to the Drupal Extension, these could go into a dedicated EntityContext class so people could enable that as needed."

Questions:
1) Is "Given (a/an) :entity entity/entities/item/items" the only relevant step definition for basic entities?
2) Is the turnip syntax I'm suggesting - requiring the keyword "entity" or "item" - the right one?
3) Am I correct that implementation would require
- a nakedEntityCreate and createNakedEntity corresponding to every nodeCreate and createNode that there is in the Drupal Driver (D7 & D8) and Drupal Extension;
- likewise for other "node" syntax like nodeDelete, beforeNodeCreate, etc.
- a step definition using the syntax above
- a feature to test this module is working right

Lastly, teardown is a significant challenge. AFAIK this project removes test data by removing all data created by the user created in the test process. However, basic entities do not necessarily record who created or owns them. My suggestion is that when working with basic entities we would have to record the full set of fields specified in the Given, and then delete entities that match those fields. 

This could potentially delete too many entities in the case that the supplied fields overlapped with existing entities of the type, so we would need to document clearly the importance of giving novel values for the fields of the entities to be Given...
"Given a Pizza item with Flavor "test flavor gdfdfsh"

jonathanjfshaw:
http://cgit.drupalcode.org/lightning/tree/tests/features/bootstrap/entity.behat.inc?h=8.x-1.x
Looks like this functionality has been created for the Drupal Lightning distribution. Hopefully it's ripe for picking out and including here.

djdevin:
Basic subcontext I wrote up that does the basics.

Understandable it needs to be abstracted for D7/D8. I named the steps similar to how they are for content in Drupalextension.
```
<?php

/**
 * @file
 * Contains \EntitySubContext.
 */
use Drupal\DrupalExtension\Context\RawDrupalContext;
use Drupal\DrupalExtension\Context\DrupalContext;
use Drupal\DrupalExtension\Context\MinkContext;
use Behat\Gherkin\Node\TableNode;

/**
 * Subcontext for creating and cleaning up entities of any type.
 */
class EntitySubContext extends RawDrupalContext {

  /** @var DrupalContext */
  private $drupalContext;

[...]

PR #300, opened 2016-08-07
Create EntityContext

jonathanjfshaw:
Basic features work locally, let's see what Travis says about the before/after hooks.


jonathanjfshaw:
Fixes #194.

jonathanjfshaw:
Relies on parallel PR https://github.com/jhedstrom/DrupalDriver/pull/101

djdevin:
Heh, I think it's only failing because it's checking Drupal.org changed their layout and some of the tests are checking regions. The new design launched in September but those tests are likely from before then. Core drupalextension is failing too. https://travis-ci.org/jhedstrom/drupalextension

Looks good though, I tested this PR against https://github.com/jhedstrom/DrupalDriver/pull/113/ and was able to add entities, view them, and have them cleaned up in D7.

jonathanjfshaw:
Great, very good to know.
#310 is the issue to fix the tests for drupalextension to resolve the drupal.org issue.

jonathanjfshaw:
We need to add make sure the tests this PR adds are tagged for D7 and pass with your PR. 

@djdevin Could you try the new tests from this PR make a PR to my repo (source of this PR) with any needed changes?

features/entity.feature is the file; I suggest you add the @d7 tag to any test that should pass in D7. You may want to add/modify some tests if some of the current ones are fundamentally incompatible with D7. 

Some of the tests use a comment type defined in fixtures/drupal8/modules/behat_test/config/install/comment.type.user_comments.yml. That will obviously fail unless an equivalent D7 fixture is created.

jonathanjfshaw:
I thought these should pass now the blackbox tests have been fixed, but there is a failing test. 

The FeatureContext function tagged @beforeEntityCreate hook is not firing - the whole function is not being invoked - and so the scenario I created to test it fails. I can't figure out the cause; it needs someone with either a better understanding of the Behat hook system or a better debugging setup than I have.

I couldn't copy the Node/User/Term hooks exactly because they extend an already existing EntityScope, so I had to name mine OtherEntityScope; perhaps somewhere here lies the root of the problem. In particular, I don't know if it matters what constants are used:
```
abstract class OtherEntityScope extends BaseEntityScope {

  const BEFORE = 'otherentity.create.before';
  const AFTER = 'otherentity.create.after';

}
```

Perhaps @jhedstrom or @pfrenssen can give a pointer.

features/bootstrap/FeatureContext.php
```diff
@@ -67,6 +67,22 @@ public function alterUserParameters(EntityScope $scope) {
...
+     */
+    public function alterEntityParameters(EntityScope $scope)
+    {
+        // @see `features/entity.feature`
+        // Concatenate 'First name' and 'Last name' to form user name.
+        $entity = $scope->getEntity();
+        if (isset($entity->{"First name"}) && isset($entity->{"Last name"})) {
+            $entity->name = $entity->{"First name"} . ' ' . $entity->{"Last name"};
+            unset($entity->{"First name"}, $entity->{"Last name"});
+        }
```
pfrenssen: This code only works for nodes, not for all entities. The code in `FeatureContext` is meant as an example to show people how things can be implemented so we should demonstrate best practices.

Instead of checking if certain parameters are set on the entity, we should check its type, I would love this code to look like this:

```
if ($entity->bundle() === 'node') {
  $entity->name = ...;
  // ...
}
```

This means however that we have to transform `$entity` from a `stdClass` into a real class. I think this would be a good thing though, and something that would be great to have in 4.x.

src/Drupal/DrupalExtension/Context/EntityContext.php
```diff
@@ -0,0 +1,89 @@
...
+  public static function getTranslationResources()
+  {
+    return glob(__DIR__ . '/../../../../i18n/*.xliff');
+  }
+
+  /**
+   * @Given I am viewing a/an :entity_type entity/item with the :label_name :label
+   * @Given I am viewing a/an :bundle :entity_type entity/item with the :label_name :label
+   */
+  public function createEntity($entity_type, $bundle = NULL, $label_name, $label)
```
pfrenssen: I don't like this from a BDD perspective, this is using the machine name of the label property in the step definition. The machine names are not usable for several reasons: first of all they often don't match the domain specific language, they might contain abbreviations and special symbols like underscores, and they are usually in English so including machine names breaks translatability.

Also, not all entities have a label, this is not a required base field. I just had a look and actually it seems like entities with labels are in the minority, most do not have one. We cannot determine it programmatically either because there is no entity key for the label.

I would drop this step.

src/Drupal/DrupalExtension/Context/EntityContext.php
```diff
@@ -0,0 +1,89 @@
...
+   * @Given :entity_type entities/items:
+   * @Given a/an :bundle :entity_type entity/item:
+   * @Given :bundle :entity_type entities/items:
+   */
+   public function createEntities($entity_type, $bundle = NULL, TableNode $entitiesTable)
+   {
+     foreach ($entitiesTable->getHash() as $entityHash) {
+       $entity = (object)$entityHash;
+       if (!isset($entity->step_bundle)) $entity->step_bundle = $bundle;
+       $this->entityCreate($entity_type, $entity);
```
pfrenssen: I guess you've chosen `step_bundle` to avoid potential conflicts with actual properties like `type` and `bundle` on real entities? This is another argument for using a real class for the `$entity` instead of a plain object.

src/Drupal/DrupalExtension/Context/EntityContext.php
```diff
@@ -0,0 +1,89 @@
...
+   * | title     | My entity        |
+   * | Field One | My field value |
+   * | author    | Joe Editor     |
+   * | status    | 1              |
+   * | ...       | ...            |
+   *
+   * @Given I am viewing a/an :entity_type entity/item:
+   * @Given I am viewing a/an :bundle :entity_type entity/item:
+   */
+  public function assertViewingEntity($entity_type, $bundle = NULL, TableNode $fields)
```
pfrenssen: I think this is making too many assumptions again, not all entities have a canonical path where it can be viewed. For example this cannot be used with `File`, `BlockContent` or `ContentModerationState` entities. I would drop this, the added benefit of directly going to the canonical path doesn't weigh up against the confusion that people will have if their test fails unexpectedly.

I think we only need a single step, to create an entity of a given type and bundle.

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -603,4 +611,39 @@ public function loggedInWithRole($role) {
...
+   * Create an entity.
+   *
+   * @return object
+   *   The created entity.
+   */
+  public function entityCreate($entity_type, $entity) {
+	if (empty($entity_type)) {
+	        throw new \Exception("Steps must specify an entity type in order to create an entity.");
+	}
+    // We want hooks & cleanup to know the entity type
```
pfrenssen: Tabs used instead of spaces for indentation, I also saw this in some other places.

features/bootstrap/FeatureContext.php
Berdir: This has nothing to do with nodes, nodes don't have a name field. This is a user. also, the method would be getEntityTypeId(), not bundle. bundle is article, page, ...

jonathanjfshaw: Yes, it would be great if these entity hooks could know what entity type they are being passed. That could be achieved by a "step_type" property, the same awful hack I'm using to keep track of bundles using "step_bundle" without colliding with entities that have fields of their own called "bundle" or "type". I agree we should create a real object to avoid these awful workarounds.

 However, this "step_bundle" syntax has already been committed on the Drupal Driver, and the code is usable without adding a "step_type" argument. I suggest we accept this PR as is, and then rearchitect to use a real class in a subsequent PR as that naturally has consequences/uses beyond this EntityContext. (Although that could be a BC break for anyone who wrote hooks or cleanup code expecting step_bundle to be there). I've opened #337 to discuss the bigger issues here.

jonathanjfshaw:
I'm currently treating this as postponed on #337 .

jonathanjfshaw:
Rebased

brooke-heaton:
This would seem like a critical improvement - we can't yet create Media entities in Drupal 8 with the Extension. Anything we are waiting on here?

noemi-mancini:
Agree with @brooke-heaton .

We have used this patch extensively on several D8/D9 projects.
With the update to D10 and the module update to v5.0.0rc1, the patch no longer applies, impacting many of our behat tests.
Is there a chance the patch will be integrated, or should we bring it in as custom code?

AlexSkrypnyk:
@jonathanjfshaw 
Thank you for this PR.

We are reviewing and closing outdated issues.

Please feel free to rebase and re-open this PR if this is still relevant.

Thank you

PR #302, opened 2016-08-09
In step to select Radio button, look for label explicitly by ID instead of parent()

jonpugh:
When this step does a final compare of the actual text of the label to the label parameter the test entered, it does a lookup on ->parent().  This fails the test if the radio button has a description, in drupal this can happen a lot.

This PR replaces the parent() call to a lookup on the element with the "for" attribute pointed at the ID of the radio element.


jhedstrom:
I've rerolled this in #437. Thanks for the fix, and sorry for the long delay!

PR #305, opened 2016-08-17
instruct to use composer global require command

jcandan:


jhedstrom:
This could be re-worked to optionally do this, but all projects I've worked on use non-global installs of the Behat libraries, so I wouldn't want to remove that bit of documentation.

jhedstrom:
Closing out for now. Feel free to re-open for further discussion. Thanks!

Issue #306, opened 2016-08-23
Allow to share information about the current user between different contexts


pfrenssen:
Every individual context extends their own instance of `RawDrupalContext`, which means that the user which is logged in in one particular instance is not available in other contexts.

An example: in our project we have a step definition that is defined in a `OgSubContext` context that allows a user to log in as a member of a certain organic group:

> Given I am logged in as a member of the :group group

This works similar to `DrupalContext::assertAuthenticatedByRole()` but will make sure that the user that is created is also subscribed to the given organic group.

Calling `$this->login()` in `OgSubContext` sets the user object in `OgSubContext::$user` but this is local to that instance. All other instances do not have access to this. If for example a step definition in `FeatureContext` accesses `$this->user` it will turn up empty.

There is always only 1 single logged in user in a given Behat test, so we should make the current user shareable between all contexts.

pfrenssen:
This has been implemented in #307. This issue can be closed.

PR #307, opened 2016-08-23
Make the current user available in all contexts.

pfrenssen:
This is a proof of concept for #306.

I have created a very simple `DrupalUserManager` service that has a getter and setter for the current user object. This allows to persist the logged in user and share it between the different contexts that all extend `RawDrupalContext`.

In order to make this work with existing code that accesses the `RawDrupalContext::$user` property directly I'm hijjacking it with a magic getter and setter. This should in theory be compatible with all  existing 3.x code, but let's have a look at how this affects the current test suite first :)

Now that we have a 4.x branch I'd love to go a bit further. We can get rid of all code that directly accesses `RawDrupalContext::$user` and `RawDrupalContext::$users`, and replace it with this service.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/DrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/DrupalContext.php
@@ -110,12 +110,10 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext {
    * @Given I am logged in as :name
    */
   public function assertLoggedInByName($name) {
-    if (!isset($this->users[$name])) {
-      throw new \Exception(sprintf('No user with %s name is registered with the driver.', $name));
-    }
+    $manager = $this->getUserManager();
 
     // Change internal current user.
-    $this->user = $this->users[$name];
+    $manager->setCurrentUser($manager->getUser($name));
 
     // Login.
     $this->login();
@@ -247,7 +245,7 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext {
    * @Given I am viewing my :type (content )with the title :title
    */
   public function createMyNode($type, $title) {
-    if (!isset($this->user->uid)) {
+    if ($this->getUserManager()->currentUserIsAnonymous()) {
       throw new \Exception(sprintf('There is no current logged in user to create a node for.'));
     }
 
@@ -255,7 +253,7 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext {
       'title' => $title,
       'type' => $type,
       'body' => $this->getRandom()->name(255),
-      'uid' => $this->user->uid,
+      'uid' => $this->getUserManager()->getCurrentUser()->uid,
     );
     $saved = $this->nodeCreate($node);
 
--- a/src/Drupal/DrupalExtension/Context/DrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/DrupalContext.php
@@ -61,7 +61,7 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext {
       }
 
       // Login.
-      $this->login();
+      $this->login($user);
     }
   }
 
@@ -101,7 +101,7 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext {
       }
 
       // Login.
-      $this->login();
+      $this->login($user);
     }
   }
 
@@ -116,7 +116,7 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext {
     $manager->setCurrentUser($manager->getUser($name));
 
     // Login.
-    $this->login();
+    $this->login($manager->getUser($name));
   }
 
   /**
@@ -138,7 +138,7 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext {
     $this->roles[] = $rid;
 
     // Login.
-    $this->login();
+    $this->login($user);
   }
 
   /**
--- a/src/Drupal/DrupalExtension/Context/DrupalSubContextBase.php
+++ b/src/Drupal/DrupalExtension/Context/DrupalSubContextBase.php
@@ -34,15 +34,22 @@ abstract class DrupalSubContextBase extends RawDrupalContext implements DrupalSu
 
   /**
    * Get the currently logged in user from DrupalContext.
+   *
+   * @deprecated
+   *   Deprecated in 4.x, will be removed before 5.x.
+   *   The currently logged in user is now available in all context classes.
+   *   Use $this->getUserManager()->getCurrentUser() instead.
    */
   protected function getUser() {
-    /** @var DrupalContext $context */
-    $context = $this->getContext('\Drupal\DrupalExtension\Context\DrupalContext');
-    if (empty($context->user)) {
+    trigger_error('DrupalSubContextBase::getUser() is deprecated. Use RawDrupalContext::getUserManager()->getCurrentUser() instead.', E_USER_DEPRECATED);
+
+    $user = $this->getUserManager()->getCurrentUser();
+
+    if (empty($user)) {
       throw new \Exception('No user is logged in.');
     }
 
-    return $context->user;
+    return $user;
   }
 
   /**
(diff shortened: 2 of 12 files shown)


pfrenssen:
Awesome, as far as I can see all tests are green, there are only failures because I didn't update PHPSpec with the new `DrupalUserManager` service that is injected in `DrupalDriverManager`.

I'll try to update that, I don't have a lot of experience with PHPSpec.

jhedstrom:
I really like this idea. It will make life easier on folks, as I think created users are the most often needed between various contexts.

Note that this is technically possible already, but all contexts needed must be manually gathered, and can only be done from custom contexts:

```php
  /**
   * @BeforeScenario
   */
  public function gatherContexts(BeforeScenarioScope $scope) {
    $environment = $scope->getEnvironment();

    $this->drupalContext = $environment->getContext('Drupal\DrupalExtension\Context\DrupalContext');
    $this->minkContext = $environment->getContext('Drupal\DrupalExtension\Context\MinkContext');
  }
```

This is rather obscurely documented [here](https://behat-drupal-extension.readthedocs.io/en/3.1/subcontexts.html).

pfrenssen:
That's a great feature if you know in advance which context was used to create the user :)

We were using `DrupalSubContextBase::getContext()` which also uses this `->getEnvironment()->getContext()` trick, but it all fell apart when we started to create and log in users in our `OgSubContext`. Loading the `DrupalContext` came up with an empty user because it was never defined there.

For continuing on this, which approach would you prefer?

* This backwards compatible approach with the magic getter and setter seems to work fine for the current 3.x branch. For 4.x we should replace the property completely with a set of methods so we don't need the magic methods. Would you like to have this in 3.x, or would this be a 4.x only feature?
* Do you like having a separate `DrupalUserManager` service, or should this functionality be included in `DrupalDriverManager`?
* How shall we deal with the `RawDrupalContext::$users` array that keeps track of all users that are created during the test scenario so they can be cleaned up? I was thinking for 3.x it would be fine to keep the current behavior: each context keeps track of its own users, and cleans them up. For 4.x we could have more dedicated methods, like `DrupalUserManager::addUser()`, `DrupalUserManager::deleteUser()` and `DrupalUserManager::getUsers()`.

pfrenssen:
I'm going to continue on this today. I'll do this for now:

* I will work on the 4.x path, replacing the direct access of the `RawDrupalContext::$user` and `::$users` properties with method calls. I will keep the magic getter and setter in place but will trigger a deprecated warning so people moving from 3.x -> 4.x will be notified that they need to update their code.
* I will keep the `DrupalUserManager` service, I think it makes sense this is encapsulated in its own object. The overhead of having this additional service is negligible.
* I will provide some additional methods to keep track of created users.

jhedstrom:
The approach outlined in your last comment makes sense to me.

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -451,22 +503,23 @@ public function languageCreate(\stdClass $language) {
...
 
   /**
-   * Log-in the current user.
+   * Log-in the given user.
+   *
+   * @param \stdClass $user
+   *   The user to log in.
    */
-  public function login() {
+  public function login(\stdClass $user) {
```
pfrenssen: I changed the way this works because the way this used to work briefly left the user in an inconsistent state. We used to first populate the `$this->user` property, and then call `$this->login()`. This meant the "current user" was already set before the user was even logged in.

It seems more logical and consistent to me to actually set the current user after the user has been successfully logged in.

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -476,20 +529,23 @@ public function login() {
...
+
+    $manager->setCurrentUser($user);
   }
 
   /**
    * Logs the current user out.
    */
   public function logout() {
     $this->getSession()->visit($this->locatePath('/user/logout'));
+    $this->getUserManager()->setCurrentUser(FALSE);
```
pfrenssen: I added this change because this seems to have been missing before. If we log the user out through the UI we should also clear the current user.

src/Drupal/DrupalExtension/Context/RawDrupalContext.php
```diff
@@ -524,7 +580,14 @@ public function loggedIn() {
...
     // perfect, this is how Drupal SimpleTests currently work as well.
-    return $page->findLink($this->getDrupalText('log_out'));
+    if ($page->findLink($this->getDrupalText('log_out'))) {
+      return TRUE;
+    }
+
+    // The user appears to be anonymous. Clear the current user from the user
+    // manager so this reflects the actual situation.
+    $this->getUserManager()->setCurrentUser(FALSE);
+    return FALSE;
```
pfrenssen: This addition should help to keep the current user in the context in sync with the actual state in the website. It is always possible the user has been logged out without by performing certain actions in the website - without the context having knowledge of this fact.

Now, if we detect the user is actually logged out in the site, the current user is updated accordingly.

src/Drupal/DrupalExtension/Context/DrupalContext.php
```diff
@@ -110,15 +110,13 @@ public function assertAuthenticatedByRoleWithGivenFields($role, TableNode $field
    * @Given I am logged in as :name
    */
   public function assertLoggedInByName($name) {
-    if (!isset($this->users[$name])) {
-      throw new \Exception(sprintf('No user with %s name is registered with the driver.', $name));
-    }
+    $manager = $this->getUserManager();
```
pfrenssen: This exception has moved to the respective method on the user manager.

src/Drupal/DrupalExtension/Context/DrupalContext.php
```diff
@@ -247,15 +245,15 @@ public function createNode($type, $title) {
    * @Given I am viewing my :type (content )with the title :title
    */
   public function createMyNode($type, $title) {
-    if (!isset($this->user->uid)) {
+    if ($this->getUserManager()->currentUserIsAnonymous()) {
```
pfrenssen: I like that as a side effect this also improves readability :)

pfrenssen:
Made the proposed changes. Am pretty happy with how this turned out. This is mostly a refactoring so the existing test coverage should be pretty good to indicate whether it works as intended.

I would like to still add a test though that logs in a user in a separate context so that we have proof that the sharing of users between contexts works as intended. This will be for tomorrow though.

pfrenssen:
I have added a test. I have also created a branch containing only the test, but not the fix. This is expected to fail. See #309.

pfrenssen:
This is ready for review. The blackbox tests are failing but these are due to #205 - I think something changed on drupal.org again.

src/Drupal/DrupalExtension/Context/DrupalSubContextBase.php
```diff
@@ -34,15 +34,22 @@ public function __construct(DrupalDriverManager $drupal) {
...
+   * @deprecated
+   *   Deprecated in 4.x, will be removed before 5.x.
+   *   The currently logged in user is now available in all context classes.
+   *   Use $this->getUserManager()->getCurrentUser() instead.
    */
   protected function getUser() {
-    /** @var DrupalContext $context */
-    $context = $this->getContext('\Drupal\DrupalExtension\Context\DrupalContext');
-    if (empty($context->user)) {
+    trigger_error('DrupalSubContextBase::getUser() is deprecated. Use RawDrupalContext::getUserManager()->getCurrentUser() instead.', E_USER_DEPRECATED);
```
jhedstrom: This is great!

jhedstrom:
I opened #310 for the long term, but will try to fix the blackbox tests in the short term.

jhedstrom:
And #311 for the short term fix.

pfrenssen:
Thanks for opening those issues. So this is now blocked by #311. I will merge #311 in here, hopefully that will turn the tests green.

jhedstrom:
This is great. How to access the user from other contexts is one of the most frequently asked questions too, so this will fix that. Thanks!

PR #309, opened 2016-08-25
Test for 307 - expected to fail - do not merge

pfrenssen:
This is a branch that contains only the test for #307. If the test is correctly written this is expected to fail here, but succeed in #307 so it proves that the reported bug is both real, and fixed :)

### DIFF
--- a/features/d8.feature
+++ b/features/d8.feature
@@ -30,3 +30,10 @@ Feature: DrupalContext
     Given I am not logged in
     When I am on the homepage
     Then I should see the heading "Search" in the "left sidebar" region
+
+  # This tests that a user that is created in one particular Context class (in
+  # this case FeatureContext::assertLoggedInByUsernameAndPassword()) can be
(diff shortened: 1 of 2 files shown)


features/d8.feature
```diff
@@ -30,3 +30,10 @@ Feature: DrupalContext
     Given I am not logged in
     When I am on the homepage
     Then I should see the heading "Search" in the "left sidebar" region
+
+  # This tests that a user that is created in one particular Context class (in
+  # this case FeatureContext::assertLoggedInByUsernameAndPassword()) can be
+  # accessed in another Context (DrupalContext::assertLoggedInByName()).
+  Scenario: Logging in as a user without an e-mail address.
+    Given I am logged in as a user with name "Carrot Ironfoundersson" and password "citywatch1234"
+    Then I am logged in as "Carrot Ironfoundersson"
```
jhedstrom: I think all of our test users should have this name!

pfrenssen:
This has been committed as part of #307. This can be closed.

PR #311, opened 2016-08-25
Fix blackbox tests

jhedstrom:


pfrenssen:
This looks good as a short term fix. The pages that we tested on drupal.org have been updated, and this reflects the current state of those pages.

jhedstrom:
Merged to 3.2 as well.

Issue #313, opened 2016-09-02
Support creating and placing blocks


acbramley:
It would be great to have step definitions similar to nodes, taxonomy terms, etc for blocks!

You could also use the config step code to set the region/weight/visibility as well

jhedstrom:
Good idea! If somebody finds themselves working on this, I think the way to go would be to add a new `BlockContext` class.

subhojit777:
I guess we have to handle the block placing from the DrupalDriver itself. BlockContext will just use the method to place the block.

PR #315, opened 2016-09-27
Issue #313: Step definition for placing block

subhojit777:
This is just the start, let me know if I am going in the right direction.

This is only for Drupal 7


jhedstrom:
@subhojit777 this looks like a good start.

One thought though, is that since this is using the admin UI to place the blocks, it ties it (as you mention) directly to Drupal 7. I would prefer adding the place block method to the https://github.com/jhedstrom/DrupalDriver, and then this new step definition need only call a method, rather than placing the blocks through the UI. This would have the added benefit of supporting more than one version of Drupal.

Once that is done, we can add a new test in `features` to verify the behavior.

subhojit777:
But using DrupalDriver we won't be able to access the UI. I haven't tried this before, but we might be able to place the block by querying the `block` table (atleast in Drupal 7) - may be that can work. But the challenge is to obtain the block delta. For example, `facetapi` it generates a hash-ed block delta, and then again we have to access `facetapi`'s  API to generate the hash-ed delta. This will add dependency to the project.

subhojit777:
Using Mink inside the DrupalDriver will also defeat its purpose.

jhedstrom:
> But using DrupalDriver we won't be able to access the UI

> Using Mink inside the DrupalDriver will also defeat its purpose.

Right--from the Drupal Driver, the API itself would be used to place the block. A step definition for placing a block probably doesn't need to use the UI, unless the express purpose was to test the UI itself, in which case that would be outside the scope of this project.

PR #318, opened 2016-10-10
Changed curl typo to use -sS flag

AndraeRay:
As seen on this page https://packagist.org/packages/drupal/drupal-extension,
the curl flag should be -sS instead of -s


jhedstrom:
Thanks for this. However, given the instructions in https://getcomposer.org/download/ I think we should instead update the documentation to just link there, since as they note:

> WARNING: Please do not redistribute the install code. It will change with every version of the installer. Instead, please link to this page or check how to install Composer programmatically.

Issue #314, opened 2016-09-09
Taxonomy term hierarchy


kiwimind:
Having tried to create taxonomy terms with parents, as suggested in the example, I'm finding that the taxonomy terms are all added at the top level, without the parent being set.

`
/**
   * Creates one or more terms on an existing vocabulary.
   *
   * Provide term data in the following format:
   *
   * | name  | parent | description | weight | taxonomy_field_image |
   * | Snook | Fish   | Marine fish | 10     | snook-123.jpg        |
   * | ...   | ...    | ...         | ...    | ...                  |
   *
   * Only the 'name' field is required.
   *
   * @Given :vocabulary terms:
   */
`

Is this something that isn't covered and needs removing from the example, or a bug that needs looking into?

For reference, the code I'm using is: 

`
    Given "regions" terms:
      | name             | parent      |
      | Parent term      |             |
      | Child term one   | Parent term |
      | Child term two   | Parent term |
`

Thanks.

claudiu-cristea:
Fix in #322

PR #321, opened 2016-10-22
#314 Fix proposal for multiple parents

haringsrob:
Solves #314 

If someone could help me setup on how I run the features in this standalone package I can also provide the tests.


jonathanjfshaw:
Is this needed given #322 which seems much cleaner? I really don't like hardcoding entity-type specific logic into a generic function like parseEntityFields, it's building a maintenance nightmare.

PR #322, opened 2016-10-22
Be aware about taxonomy term parent, if was passed

claudiu-cristea:
This is a fix for #314.


jhedstrom:
Can a test (in `api.feature`) be added that uses this new parent functionality?

jhedstrom:
It also needs to be rebased or merged with latest `master`.

Issue #310, opened 2016-08-25
Do not use Drupal.org for blackbox tests


jhedstrom:
The blackbox tests have been using https://www.drupal.org since this project has existed. Many times they break due to a change, which causes all PRs to fail.

For 4.0, we should replace these with a simple static HTML  site in our test fixtures directory.

PR #324, opened 2016-11-10
Revert to stable release of Behat

pfrenssen:
I just noticed in our project we were stuck on an old release of Behat, while the new 3.2.2 version was released a couple of days ago. Seems like we forgot to update our dependencies when 3.1.0 was released back in March.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,11 +15,11 @@
     "behat/mink": "~1.5",
     "behat/mink-goutte-driver": "~1.0",
     "behat/mink-selenium2-driver": "~1.1",
-    "behat/behat": "~3.1.0-rc2",
+    "behat/behat": "~3.2",
     "behat/mink-extension": "~2.0",
     "drupal/drupal-driver": "dev-master",
-    "symfony/dependency-injection": "~2.7",
-    "symfony/event-dispatcher": "~2.7"
+    "symfony/dependency-injection": "~2.7|~3.0",
+    "symfony/event-dispatcher": "~2.7|~3.0"
   },
   "require-dev": {
     "phpspec/phpspec": "~2.0",
--- a/README.md
+++ b/README.md
@@ -84,3 +84,32 @@ the [Full documentation](https://behat-drupal-extension.readthedocs.org)
  * [Drupal form element visibility](https://gist.github.com/pbuyle/7698675)
  * [Track down PHP notices](https://www.godel.com.au/blog/use-behat-track-down-php-notices-they-take-over-your-drupal-site-forever)
  * [Support for sites using basic HTTP authentication](https://gist.github.com/jhedstrom/5bc5192d6dacbf8cc459)
+
+## Release notes
+
+### Backwards incompatible changes
+
+Starting with 3.3.0 Behat Drupal Extension depends on Behat 3.2.0 which
+requires all callbacks to be defined as static methods.
+
+Before 3.3.0:
+
+```
+/**
+ * @afterUserCreate
+ */
+public function afterUserCreate(EntityScope $scope) {
+  // ...
+}
+```
+
+Starting with 3.3.0:
+
+```
+/**
+ * @afterUserCreate
+ */
+public static function afterUserCreate(EntityScope $scope) {
+  // ...
+}
+```
--- a/features/bootstrap/FeatureContext.php
+++ b/features/bootstrap/FeatureContext.php
@@ -21,8 +21,8 @@ class FeatureContext extends RawDrupalContext {
    *
    * @beforeNodeCreate
    */
-  public function alterNodeParameters(BeforeNodeCreateScope $scope) {
-    parent::alterNodeParameters($scope);
+  public static function alterNodeParameters(BeforeNodeCreateScope $scope) {
+    call_user_func('parent::alterNodeParameters', $scope);
     // @see `features/api.feature`
     // Change 'published on' to the expected 'created'.
     $node = $scope->getEntity();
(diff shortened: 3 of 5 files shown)


pfrenssen:
This would probably also be a good candidate to backport to the 3.2 branch.

pfrenssen:
Forget about the backporting, we have some massive failures it seems :)

jhedstrom:
Looks like the vast majority of the failures are in the blackbox feature, which relies on drupal.org not changing (which it has). Might be as good a time as any to have a simple html fixture 'site' that the blackbox test runs against...

pfrenssen:
Good plan, the issue about stabilizing the blackbox tests is #310

pfrenssen:
The blackbox tests have been fixed, let's try this again.

pfrenssen:
This is green here but this needs some discussion before it can be merged.

Somewhere inbetween the old 3.1.0-rc2 version we were using and the current stable release of Behat a requirement was added to make all callbacks static methods. This makes sense from an architectural standpoint but this might cause backwards compatibility breaks for people that have made their own custom callbacks.

I tried this on our big D8 project and we had a handful of instances where we needed to update some code (ref. https://github.com/ec-europa/joinup-dev/pull/323).

It's hard to predict how many people will be affected by this, and there is a very simple fix to work around it:

```
$ composer require --dev behat/behat:~3.1.0-rc2
```

There has been a report recently by @Trogie that he was affected by our usage of the old version (ref #332).

We have two options:

1. Only apply this fix in the 4.x branch, nobody who is using 3.x will be affected, but they will not be able to use newer versions of Behat.
2. Apply the fix in 3.x, and provide instructions in the README and release notes how to work around the issue. It is quite easy to fix for any developer worth its salt, but it is likely that people will not read the README or release notes.

@jhedstrom, @Trogie, your thoughts?

jhedstrom:
I prefer approach 2, where we commit to 3.x with instructions.

jhedstrom:
We could in theory cut a 3.3 branch, so it's less disruptive...

pfrenssen:
Ok great, that was also my preference :)

There is some more work required here though, I am seeing that in my project my `@beforeCreateNode` hooks are executed multiple times using the newest version of Behat.

pfrenssen:
I tracked it down to https://github.com/Behat/Behat/issues/758 which is included in Behat 3.2.0. I have a workaround.

pfrenssen:
While debugging this I noticed that the way the subcontexts are discovered is not optimal. I created a followup for this: #340. We can also look to get rid of the workaround that is added here in that issue.

pfrenssen:
I did a new test run on our D8 platform which has 2513 steps in 229 scenarios, and it all passes now, after declaring our callbacks as static methods (ref. https://github.com/ec-europa/joinup-dev/pull/323).

pfrenssen:
I have also run a test on a D7 platform with 1149 steps in 69 scenarios and it ran fine. I had to make an adaptation because one of the steps was using a workaround to get the current user from a context class, which is now solved by the `UserManager` service in the master branch. This is something that would not happen once this is backported to 3.x.

Ref. https://github.com/ec-europa/platform-dev/pull/1131

Trogie:
Whatever solution is best for all! :D

pfrenssen:
I will rebase this branch so it is one single commit, that will be the easiest to backport to 3.x. I think this probably can go in the master branch right away.

@jhedstrom I think a 3.3 release would indeed be least disruptive. Have you got a timeline in mind it? 3.2 is not that old yet. We could also support 3.2 and 3.3 in parallel for a while, I see this also was done in the past for the 3.0 branch.

composer.json
```diff
@@ -15,7 +15,7 @@
     "behat/mink": "~1.5",
     "behat/mink-goutte-driver": "~1.0",
     "behat/mink-selenium2-driver": "~1.1",
-    "behat/behat": "~3.1.0-rc2",
+    "behat/behat": "~3.2",
     "behat/mink-extension": "~2.0",
     "drupal/drupal-driver": "dev-master",
     "symfony/dependency-injection": "~2.7",
```
jhedstrom: We should remove the symfony verison pinning here too if possible. See #372.

siwinski:
Any idea when this will get merged?  I am setting up a new Behat test suite and I currently have to choose between using either an older version of Behat or using an older version of this extension ([version 3.1.5 allows `"behat/behat": "~3.0,>=3.0.5"`](https://github.com/jhedstrom/drupalextension/blob/v3.1.5/composer.json#L18)).

kmajzlik:
I have no idea how to use Drupal 8.4-dev with Behat now... Can somebody review and merge please?

haringsrob:
Hi @jhedstrom,

I have updated @pfrenssen 's branch to use 3.x versions of symfony as well.

Lets hope all is green.

pfrenssen:
Thanks @haringsrob for getting the ball rolling again!

If this blocks the usage of Drupal 8.4.x then I think this would be a good moment to make a 3.3.0 release, @jhedstrom do you agree?

haringsrob:
First failures are coming in,

I think it is ok to drop support for EOL php versions..

Currently php 5.6 and up are still supported: http://php.net/supported-versions.php

This would also be a good message towards developers/devops to start upgrading their systems.

Same goes for Drupal 6, we can keep it in, but support should no longer be maintained.

pfrenssen:
There is a different PR for the EOL PHP versions: #335.

I'm fine with dropping Drupal 6 too, it's been deprecated for a long time and we are moving to a new version so a B/C break is fine. This definitely needs approval from @jhedstrom though.

We can keep maintaining 3.2.x for some time for people needing D6.

pfrenssen:
Drupal 6 still works for PHP 5.5, let's keep it in for now. We can drop Drupal 6 when we go to 4.0.0.

Dropping PHP 5.3 and 5.4 are fine I think, especially if we keep maintaining 3.2.x for the time being.

pfrenssen:
OK #335 and #381 are in. I rebased this on top of the latest master. Let's see how it goes.

jhedstrom:
I'm good with making a `3.3` release once this is in. Thanks @haringsrob for helping out!

pfrenssen:
And it's green! Great!

PR #327, opened 2016-11-22
Use a more specific exception to indicate unmet expectations in MessageContext

pfrenssen:
I am using an `@AfterStep` to take screenshot if I have PHP notices on the screen. I suddenly noticed this started taking screenshots of every step, even if there were no PHP notices.

This was because I was checking the notices with the MessageContext, and somewhere in the chain an exception was thrown. My code could not distinguish between an actual failed message and an unrelated exception:

```
try {
  $context->assertNotWarningMessage('Notice:');
}
catch (\Exception $e) {
  $this->saveScreenshot();
}
```

Let's use a more specific exception to indicate that our expectation was not met.

Note that this is a potential B/C break with the current code base in the 3.x branch. If people have been using code like in this example snippet, then they won't be able to catch messages any more.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MessageContext.php
+++ b/src/Drupal/DrupalExtension/Context/MessageContext.php
@@ -4,6 +4,7 @@ namespace Drupal\DrupalExtension\Context;
 
 use Behat\Behat\Context\TranslatableContext;
 use Behat\Gherkin\Node\TableNode;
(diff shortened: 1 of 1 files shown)


jhedstrom:
This looks good. Fails appear to be due to the known blackbox fails.

pfrenssen:
Thanks!

PR #331, opened 2016-12-14
ExpectationException incorrectly instantiated in MessageContext

claudiu-cristea:


src/Drupal/DrupalExtension/Context/MessageContext.php
```diff
@@ -261,14 +261,14 @@ private function _assert($message, $selectorId, $exceptionMsgNone, $exceptionMsg
     $selector = $this->getDrupalSelector($selectorId);
     $selectorObjects = $this->getSession()->getPage()->findAll("css", $selector);
     if (empty($selectorObjects)) {
-      throw new ExpectationException(sprintf($exceptionMsgNone, $this->getSession()->getCurrentUrl()));
+      throw new ExpectationException(sprintf($exceptionMsgNone, $this->getSession()->getCurrentUrl()), $this->getSession());
```
pfrenssen: Great catch! Could you make a small modification though? I had a look at the documentation of `ExpectationException` and it actually expects a `DriverInterface` object. The `Session` is only accepted for backwards compatibility reasons and is deprecated. You can get the driver from the session with `$this->getSession()->getDriver()`.

pfrenssen:
Thanks for the PR! Can you make the small modification to all three instances where the exception is instantiated?

pfrenssen:
Great, thanks!

PR #335, opened 2017-01-01
Stop testing on PHP 5.3

pfrenssen:
PHP 5.3 has become unsupported years ago. Our test suite has been failing on the 5.3 tests for over a year now. This is slowing down the test runs.

Instead of spending time trying to fix it, let's just remove it from the matrix and move on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.3
   - 5.4
   - 5.5
   - 5.6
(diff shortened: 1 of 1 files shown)


jhedstrom:
This looks good to me once the conflicts are resolved.

haringsrob:
Resolved the conflict, this needs a merge before #324

pfrenssen:
Thanks @haringsrob!

pfrenssen:
Once this is in, let's also make a PR to drop PHP 5.4, that also needs to be done before we can merge #324.

PR #347, opened 2017-01-30
BeforeStep and AfterStep no longer support tags in Behat 3

### LINKED ISSUE #92: BeforeStep and AfterStep no longer support tags in Behat 3
Tags can no longer be used on BeforeStep and AfterStep hooks. This means that this code in MinkContext.php:

```
/**
   * For javascript enabled scenarios, always wait for AJAX after clicking.
   *
   * @AfterStep @javascript
   */
  public function afterJavascriptStep($event) {
    $text = $event->getStep()->getText();
    if (preg_match('/(follow|press|click|submit)/i', $text)) {
      $this->iWaitForAjaxToFinish();
    }
  }
```

is failing silently - the function is never called.

See https://github.com/Behat/Behat/issues/650 and https://github.com/Behat/Behat/issues/653.

pfrenssen:
Fixes #92.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MinkContext.php
+++ b/src/Drupal/DrupalExtension/Context/MinkContext.php
@@ -82,9 +82,14 @@ class MinkContext extends MinkExtension implements TranslatableContext {
   /**
    * For javascript enabled scenarios, always wait for AJAX before clicking.
    *
-   * @BeforeStep @javascript
+   * @BeforeStep
    */
   public function beforeJavascriptStep($event) {
+    /** @var \Behat\Behat\Hook\Scope\BeforeStepScope $event */
+    $tags = $event->getFeature()->getTags();
+    if (!in_array('javascript', $tags)) {
+      return;
+    }
     $text = $event->getStep()->getText();
     if (preg_match('/(follow|press|click|submit)/i', $text)) {
       $this->iWaitForAjaxToFinish();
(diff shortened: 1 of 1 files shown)


pfrenssen:
Also cherry-picked to the 3.2 branch: https://github.com/jhedstrom/drupalextension/commit/c174852531195711d367444c91e94029d23ca732

jonathanjfshaw:
I'm not sure this is a good enough solution. The fix covers the case if a feature is tagged@javascript, but not if a scenario is tagged @javascript. And that difference is against usual Behat DX (in that we expect scenarios to inherit tag effects from their parent features) and is bound to catch people out.

[According to @everzet](https://github.com/Behat/Behat/issues/795) to get the expected result we have to "Use hooks and temporary state in your context (or trait). That's nasty, but will work.".

So how about:
```
  private $taggedJavascript;

    /**
    * @BeforeScenario @javascript
    */
   public function beforeScenario(BeforeScenarioScope $scope)
   {
      $this->taggedJavascript = true;
   }

   /**
    * @AfterScenario @javascript
    */
   public function afterScenario(AfterScenarioScope $scope)
   {
       $this->taggedJavascript = false;
   }

  /**
   * @BeforeStep
   */
  public function beforeJavascriptStep($event) {
    if ($this->taggedJavascript) {
      $text = $event->getStep()->getText();
      if (preg_match('/(follow|press|click|submit)/i', $text)) {
        $this->iWaitForAjaxToFinish();
      }
  }
```

pfrenssen:
I didn't know this, sounds like a good suggestion. Unfortunately this PR is already merged, would it be possible to create a new one for this?

jonathanjfshaw:
Sure. This gives us 4 javascript specific functions. Do you want to split these out into a javascript context for 4.x?

pfrenssen:
Yes a separate context for this sounds like a really good idea.

jonathanjfshaw:
We don't currently have a feature that tests this. Somewhere in the Drupal admin UI that uses AJAX could provide a test.

How about:
```
@javascript
Given I am logged in as an "administrator"
When I visit "/admin/structure/types/manage/page/form-display"
Then I should see not "Widget settings: Textfield"
When I press the "title_settings_edit" button
Then I should see "Widget settings: Textfield"
```

PR #349, opened 2017-01-31
Blanks in entity tables

### LINKED ISSUE #346: Allow entity reference fields to be optional
There are some wrinkles in declaring entities with optional entity reference fields.
For example,
(A)
```
    Given "article" content:
      | title       | field_reference    | 
      | Test1     |                            |
      | Test2     | Test1                   |
```
fails with "No entity '' of type 'node' exists." because in the first row the field_reference column is blank, which gets passed to the driver as an empty string, which doesn't match the label of any node. Things get weirder when you try to reference users and leave a row blank, because in that case a reference is created to user 0 (the anonymous user).

It is possible to workaround this by creating 2 givens:
(B)
```
    Given "article" content:
      | title       |
      | Test1     |
     Given "article" content:
      | title       | field_reference    | 
      | Test2     | Test1                   |
```
but this is inelegant.

I think it would be ideal if non-required entity reference fields could be blank as in (A), in which case no reference is created. And certainly it would be less confusing if user references behaved the same as node ones; if people want to reference the anonymous user, perhaps they should have to do that explicitly.

Implementation of this is probably a driver matter; I've started discussion here in the extension because it starts with a question of how we want step tables to be understood.

jonathanjfshaw:
Resolves #346.  Not sure this is right yet, let's see what the tests say..

### DIFF
--- a/features/field_handlers.feature
+++ b/features/field_handlers.feature
@@ -110,7 +110,7 @@ Feature: FieldHandlers
     And I should see "150-0040"
 
   @d7 @d8
-  Scenario: Test various user field handlers in Drupal 7
+  Scenario: Test various user field handlers in Drupal 7.
     Given "tags" terms:
       | name      |
       | Tag one   |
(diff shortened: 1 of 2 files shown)


pfrenssen:
Looking great! I like the test coverage too. Thanks!

pfrenssen:
I will squash this to make it easier to backport to 3.x.

pfrenssen:
Backported to the 3.2 branch: https://github.com/jhedstrom/drupalextension/commit/20fc44d303556febe2b5519aa5ce7ec315f99c96

PR #354, opened 2017-02-07
Correct --init & -dl command

kentr:
For me, the instructions didn't work as-is.

**Expected results:**

Running `bin/behat --init` & `behat -dl` would display the list of tests as the instructions say.

**Actual results:**

`bin/behat -dl` returning nothing.

### DIFF
--- a/doc/globalinstall.rst
+++ b/doc/globalinstall.rst
@@ -83,9 +83,9 @@ Set up tests
 3. Initialize behat. This creates the features folder with some basic things to
    get you started::
 
-    bin/behat --init
+    bin/behat -c path/to/behat.yml --init
 
-4. This will generate a FeatureContext.php file that looks like:
+4. This will generate a FeatureContext.php file in `test-directory/features` that looks like:
 
   .. literalinclude:: _static/snippets/FeatureContext.php.inc
      :language: php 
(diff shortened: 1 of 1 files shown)


kentr:
How's this?  

Behat is new to me, so I have only knowledge of what I encountered when I attempted to follow the installation instructions.  Wasn't sure what to do about the file path in this text: _"The `features/FeatureContext.php` file must be in the same directory as your `behat.yml`"_

Maybe the underlying issue is that the instructions probably assume that `behat` would be run inside the directory with `behat.yml`.  Is that the assumption?  

If so, my changes probably aren't the best path to take -- would probably be better to add a step like

```
cd sites/default/behat-tests
```

pfrenssen:
> Maybe the underlying issue is that the instructions probably assume that behat would be run inside the directory with behat.yml. Is that the assumption?
> 
> If so, my changes probably aren't the best path to take -- would probably be better to add a step like
> ```
> cd sites/default/behat-tests
> ```

Yes that is the assumption, it would be a good idea to update the documentation with this suggestion.

It is not a good practice to put the tests inside the web root though, so maybe just use a generic example like:

```
$ cd /path/to/behat/tests
```

jhedstrom:
For most sites in Drupal 8, the webroot will be a subdirectory of the project root (eg, `web` or `build/html`, etc), and behat will be located in a sibling subdirectory (eg, `test`), so let's make that assumption for the docs fix. Also, the binary is now in `vendor/bin/behat` since we removed `bin` from our `composer.json` file.

PR #359, opened 2017-03-01
Fix code block

Leksat:
The last code block at http://behat-drupal-extension.readthedocs.io/en/3.1/environment.html is not visible currently.


pfrenssen:
Nice catch, but this is for 3.1 which is no longer supported (hence the test failures). Any chance you could  apply this patch on the latest master branch?

jhedstrom:
I've targetted this against `master` in #433. Thanks!

PR #363, opened 2017-04-11
Update RawDrupalContext.php

zuernBernhard:
With Scenarios there the first step has to do with login/logout I get the following error: 

Given I am logged in as a user with the "XXXX" role # Drupal\DrupalExtension\Context\DrupalContext::assertAuthenticatedByRole()
      Fatal error: Call to a member function elements() on null (Behat\Testwork\Call\Exception\FatalThrowableError)

Because the Selenium2Driver is not startet at this moment. With this fix in loggedIn() it works.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
@@ -565,6 +565,12 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface {
    */
   public function loggedIn() {
     $session = $this->getSession();
+    
+    // We have to make sure that the driver is started otherwise it may not work (eg Selenium2Driver)
+    if (!$session->getDriver()->isStarted()) {
+      $session->getDriver()->start();
+    }
+    
     $page = $session->getPage();
 
     // Look for a css selector to determine if a user is logged in.


jonathanjfshaw:
I'm not sure if this is the right fix. I could imagine (with custom steps) possibilities where the session is started but no page has yet been visited. Perhaps it would be better to check that $page is not null and handle that? Seems more generic and transparent that way, what we're doing and why.

zuernBernhard:
Yes you are right it doesn't feel like the ultra-smart solution. But in fact $page is not null at this Point so I tried to implement the solution you suggested right now and it doesn't solve the problem. SO as far as we do not have a better fix, it wolud like to go for the simple check here.

jonathanjfshaw:
Hmmm, you're right about $page. But being started still doesn't guarantee we have a url and can find elements, if I understand correctly the documentation of Mink;s driver interface. 

```
    /**
     * Starts driver.
     *
     * Once started, the driver should be ready to visit a page.
     *
     * Calling any action before visiting a page is an undefined behavior.
     * The only supported method calls on a fresh driver are
     * - visit()
     * - setRequestHeader()
     * - setBasicAuth()
     * - reset()
     * - stop()
     *
     * Calling start on a started driver is an undefined behavior. Driver
     * implementations are free to handle it silently or to fail with an
     * exception.
     *
     * @throws DriverException When the driver cannot be started
     */
    public function start();
```

I wonder if it is is theoretically better to do something like:
```
try {
$url = $session->getDriver()->getCurrentUrl();
} 
catch ($e){
$url = NULL;
}
if !is_null($url) { 
//proceed to check login ...
}
```

zuernBernhard:
Here on my machine this results in 

```Fatal error: Call to a member function url() on null (Behat\Testwork\Call\Exception\FatalThrowableError)```

because the driver is not started.

**Comment by jonathanjfshaw** [comment]
So if $page is not NULL, then  presumably the exception is not coming from getPage() but from $page->has() in the following?
 ```
   // Look for a css selector to determine if a user is logged in.
    // Default is the logged-in class on the body tag.
    // Which should work with almost any theme.
    try {
      if ($page->has('css', $this->getDrupalSelector('logged_in_selector'))) {
        return TRUE;
      }
    } catch (DriverException $e) {
      // This test may fail if the driver did not load any site yet.
    }
```

If that's true, then perhaps the problem is simply that `catch (DriverException $e) ` is too specific to catch your "Exception\FatalThrowableError" and all that needs to be done here is to change it to `catch (Exception $e)`. What do you reckon?

pfrenssen:
The code in the PR will not solve this, this is doing exactly the same as the call to `Mink::getSession()` is already doing in the line preceding it.

The problem here is that on a brand new webdriver session there is no page loaded yet. In most implementations the browser will be on the `about:blank` page which doesn't have any content.

I propose to check if the page is empty, and if it is just return `FALSE` since no user will be logged in at the start of a new session.

@zuernBernhard can you check if the solution in #366 works for you?

pfrenssen:
I've been experimenting with this and I cannot replicate it. On a brand new session the `$page` is populated with an empty document and everything works as expected.

@zuernBernhard which driver are you using?

pfrenssen:
I tried `Selenium2Driver` with both Firefox and Chrome, `GoutteDriver` and `PhantomJSDriver`. I'm usign the following scenario to test:

```
Feature: User authentication

  @javascript
  Scenario: Anonymous user can see the user login page
    Given I am not logged in
```

Cannot replicate the problem with either driver.

jhedstrom:
Closing as folks cannot replicate this. Feel free to re-open if it's still an issue.

PR #377, opened 2017-05-18
Extend 'I wait for ajax' step

OmarMoper:
Hi.

This pull request include this  features to the '@Given I wait for AJAX' step:
1) Allow specify time to wait, in case requests are longer than 5 seconds, with this sintax:
Given I wait for AJAX to finish at least "5" seconds
2) Throw exception when AJAX petition didn't finish; in some case it breaks the test in another part. There is some reason to don't throw an exception in this step, maybe it's not completely trustworthy?

Can you review this changes and merge? Thanks!

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MinkContext.php
+++ b/src/Drupal/DrupalExtension/Context/MinkContext.php
@@ -116,10 +116,20 @@ class MinkContext extends MinkExtension implements TranslatableContext {
   /**
    * Wait for AJAX to finish.
    *
+   * @param int $seconds
+   *   Max time to wait for AJAX.
+   *
+   * @Given I wait for AJAX to finish at least :seconds seconds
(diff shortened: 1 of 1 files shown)


pfrenssen:
I agree that this limit is too low. Waiting for a specific number of seconds is never reliable, a test machine may be experiencing slowdowns on some occasions, causing random test failures.

It would be better if this isn't set by passing in an argument, but by a configuration value in `behat.yml`. Then if you know you are going to run a test suite on slow infrastructure you can increase the timeouts for the entire test run, or even pass in the options on the command line.

Are there any other places where we are using a hardcoded timeout?

AlexSkrypnyk:
@omarlopesino 
Thank you for this PR.

We are reviewing and closing outdated issues.

Please feel free to rebase and re-open this PR if this is still relevant.

Thank you

PR #379, opened 2017-05-25
Fix null session when first step asserts a login

deviantintegral:
```
  Scenario: Create and delete a...                         # features/mytest.feature:7
    Given I am logged in as a user with the "administrator" role # Drupal\DrupalExtension\Context\DrupalContext::assertAuthenticatedByRole()
      Behat\Testwork\Call\Exception\FatalThrowableError: Fatal error: Call to a member function elements() on null in /var/www/html/vendor/behat/mink-selenium2-driver/src/Selenium2Driver.php:488
```

https://github.com/minkphp/Mink/pull/705 removes the auto-starting of sessions as this was never a supported use of the API. This PR fixes compatibility Drupal 8.3+ which currently pulls in the master branch.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
@@ -585,6 +585,9 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface {
    */
   public function loggedIn() {
     $session = $this->getSession();
+    if (!$session->isStarted()) {
+      $session->start();
+    }
     $page = $session->getPage();
 
     // Look for a css selector to determine if a user is logged in.


pfrenssen:
This doesn't seem like the right solution. It is not guaranteed that the test starts with logging in. There might be another call to `$this->getSession()->getPage()` happening, we have over a dozen calls to this method in the code base.

pfrenssen:
I just looked at the code and this is not a bug in Behat Drupal Extension, but in Behat itself.

pfrenssen:
This is a duplicate of #366, can you check if that PR solves your problem?

deviantintegral:
I just tried with `https://github.com/jhedstrom/drupalextension/pull/366.diff` and the above error still occurs.

pfrenssen:
Ah yeah I understand, the `Session` object is not actually `NULL`, I misread the issue title.

Can you check #385?

Issue #195, opened 2015-07-27
Support for testing emails sent from Drupal


jonathanjfshaw:
Following on from https://www.drupal.org/node/1935598, planning for adding the ability to test emails sent by Drupal.

jhedstrom commented there: "The new steps should go into a dedicated context so it can be enabled as needed. I think some changes will be necessary to the Drupal Drivers too."

Currently I'm aware of 5 fleshed out available code samples on this subject, which show different approaches:
- @Berdir kindly shared his approach with me (which I've put in a google doc here https://docs.google.com/document/d/13Mw4Bc0t967bdqlCkenluCVXLKGLIa4YzoRY... for easy reference). He uses core's test_mail_collector.
- @larowlan blogs in detail here https://www.previousnext.com.au/blog/advanced-testing-drupal-emails-beha... about his approach, also using test_mail_collector.
- @douggreen (correct link is https://github.com/tag1consulting/behat-drupalextension/blob/master/src/..., lines 530-686 and 1053-1074) uses imap to check an email is actually send and received.
- @jamessharv and @jurgenhaas created a subcontext for the maillog module in #1932698
- https://github.com/NuCivic/dkanextension src/Drupal/DKANExtension/Context/MailContext.php I have not reviewed
- some StackExchange discussion here: http://drupal.stackexchange.com/questions/187643/how-do-i-inspect-mail-messages-in-behat-tests (using test_mail_collector)

I want to suggest the following approach:
- The Drupal Extension's out-of-the-box email system would use test_mail_collector. It's core not contrib (unlike maillog), and requires no setup (unlike IMAP).
- By default, testing never sends live emails. The Drupal Extension should record the original mail_system on load, swap it out for the test_mail_collector, and restore the original at the end of testing.
- A tag @live-email could be added to any feature or scenario, which would temporarily swap in the original mail_system just for that scenario; this would facilitate IMAP type solutions or manual checking, for anyone wants that
- The test_mail_collector's email cache would be flushed a) at the beginning of every feature or scenario tagged @email, and b) by the step "Given the email cache is empty"
- It's important that steps behave well even when multiple emails are sent to the same address, they don't just look at the first one

Currently used step definitions are ...

douggreen:
@Then /^I (?:(?:do|should) )?receive mail (titled )?"(?P[^"]_)"$/
@Then /^I (?:do|should) not receive mail (titled )?"(?P[^"]_)"$/
[...]

PR #392, opened 2017-07-13
Support for testing emails sent from Drupal

jonathanjfshaw:
Addresses #195.
Depends on parallel driver PR134: https://github.com/jhedstrom/DrupalDriver/pull/134.

A few architectural notes ...

## Service architecture

I had in mind the possibility that there could be fundamentally different mechanisms for handling test emails (e.g. Drupal core's test mail collector, contrib maillog, and IMAP) which could use the same context. Therefore I tried to put interacting with the mail store outside of the context as a service. The supplied implementation uses test mail collector.

However, I'm not clear yet on how do this in the Drupal extension and so it's not a real service for now, just a kind of service-like class. Hopefully though it will be relatively easy to upgrade to a real service in the future.

Abstracting in this way is responsible for a few minor complications that have no purpose in the default implementation such as 
*  the distinction between enabling/disabling email sending, and collecting/not-collecting emails. An IMAP service would want to send emails AND collect them, whereas for Drupal's test mail collector collecting always stops sending.
* the $store name passed when getting emails. Test mail collector is just one store, but different services, or combinations of services, could potentially be capturing emails into various different stores.

## Mink context
I needed to be able to visit pages in order to follow links from emails, but the context is extending RawDrupalContext not RawMinkContext. Therefore I moved getcontext() out of SubContextBase into it's parent RawDrupalContext and used this to help me load a minkContext and visit a page. Maybe there was an easier way of doing this.


jonathanjfshaw:
There's a lot of helper functions in the MailContext. It's possible that these should be split out into a RawMailContext in order to make it easier for people to override the mail steps in their own projects.

jhedstrom:
This is looking great! Thanks for working on this.

I agree that any non-step-definition code in `MailContext` should go into a `RawMailContext` to allow that to be re-used for folks that want to craft their own step-definitions.

pfrenssen:
> I needed to be able to visit pages in order to follow links from emails, but the context is extending RawDrupalContext not RawMinkContext. Therefore I moved getcontext() out of SubContextBase into it's parent RawDrupalContext and used this to help me load a minkContext and visit a page. Maybe there was an easier way of doing this.

This is also a good idea! Can we maybe split this off to a separate PR?

jonathanjfshaw:
Suggested changes by @jhestrom and @pfrenssen made. Tests are failing because this is blocked on the hunk split out into #396 and the supporting driver methods in jhedstrom/DrupalDriver#134.

jhedstrom:
Thanks for this!

Issue #386, opened 2017-06-19
getContext() method doesn't seem to exist


phenaproxima:
After updating to Drupal Extension 3.3.0, I get this error when running acquia/lightning's tests:

Fatal error: Call to undefined method Behat\Behat\Context\Environment\UninitializedContextEnvironment::getContexts() in vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/DrupalSubContextBase.php on line 64

I'm not sure what's happening here -- any ideas would be welcome.

balsama:
I traced this problem to line 64 of DrupalSubContextBase. It thinks $environment is a \Behat\Behat\Context\Environment\InitializedContextEnvironment, but it's actually \Behat\Testwork\Environment\Environment, which doesn't have the `getContexts()` method (InitializedContextEnvironment does).

pfrenssen:
I think this might be related to the Behat update that we did in the latest release. We were stuck on Behat 3.1 before. In Behat 3.2 a change was introduced that properly reinitializes Context classes at the start of every scenario and any code that relies on data being present in contexts from before the scenario started will now fail. Ref. https://github.com/Behat/Behat/issues/758

I had a look at one of the contexts that threw a failure in the build that was linked, and it seems that this is the cause of the problem:

https://github.com/acquia/lightning/blob/02b18b602fb75f0990f802fa8d60f2a59234f82b/src/LightningExtension/Context/BlockContext.behat.inc#L21

```
  /**
   * Gathers required contexts.
   *
   * @BeforeScenario
   */
  public function gatherContexts() {
    $this->minkContext = $this->getContext(MinkContext::class);
  }
```

The `@BeforeScenario` means this code runs before the scenario is started. I haven't looked at the actual implementation but it seems likely that the contexts are not yet initialized at that point so they cannot be retrieved from the environment.

This also seems like very inefficient code, since this will be executed at the start of every single scenario, even if the step definitions defined in the context are not even used in the scenario.

Can you try changing this so the contexts are retrieved when they are actually being used instead of on the `@BeforeScenario` hook? Something like this:

```
-    $this->minkContext->fillField('id', $id);
+   $this->getContext(MinkContext::class)->fillField('id', $id);
```

pfrenssen:
I have proposed a PR that throws a more helpful exception, that will hopefully in the future help developers understand & debug this problem.

Can you confirm if the `@BeforeScenario` is the root of the problem?

balsama:
[...]

PR #396, opened 2017-07-19
Move getContext() into RawDrupalContext from DrupalSubContextBase

jonathanjfshaw:
Split out of #392

### DIFF
--- a/src/Drupal/DrupalExtension/Context/DrupalSubContextBase.php
+++ b/src/Drupal/DrupalExtension/Context/DrupalSubContextBase.php
@@ -7,7 +7,6 @@
 
 namespace Drupal\DrupalExtension\Context;
 
-use Behat\Behat\Context\Environment\InitializedContextEnvironment;
 use Drupal\DrupalDriverManager;
 
 /**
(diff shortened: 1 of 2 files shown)


jonathanjfshaw:
```
// If this code is executed
// before a test scenario starts (e.g. in a `@BeforeScenario` hook) then the
// contexts cannot yet be retrieved.
```
I removed this comment as testing showed it to be untrue; `getContext()` works fine from a @BeforeScenario tagged function.

jonathanjfshaw:
This method appears to be untested. My first commit omitted the use statement and so was actually broken, but passed on Travis.

pfrenssen:
> I removed this comment as testing showed it to be untrue; getContext() works fine from a @beforescenario tagged function.

Can you check this with the latest version of Behat? This comment was added recently because it was discovered that this can no longer be used in `@BeforeScenario` in recent versions of Behat. It used to work in versions older than Behat 3.2.0.

See https://github.com/jhedstrom/drupalextension/issues/386

jonathanjfshaw:
> Can you check this with the latest version of Behat?

I tried but ended up in dependency hell. I did discover though that my testing had been on Behat 3.1, so I think we should assume your previous results are right and I restored the comment.

jhedstrom:
Merged latest master in #428.

Issue #400, opened 2017-08-04
iWaitForAjaxToFinish() not really working?


Berdir:
At least in D8, the wait for ajax to finish step isn't working and our tests are full of "And I wait 1-3 seconds", which makes them a lot slower than they have to be.

I tried an implementation based on what core uses in \Drupal\FunctionalJavascriptTests\JSWebAssert::assertWaitOnAjaxRequest, which looks a bit like this then:

```
  /**
   * Wait for AJAX to finish.
   *
   * @see \Drupal\FunctionalJavascriptTests\JSWebAssert::assertWaitOnAjaxRequest()
   *
   * @Given I wait for AJAX to finish
   */
  public function iWaitForAjaxToFinish() {
    $condition = <<<JS
    (function() {
      function isAjaxing(instance) {
        return instance && instance.ajaxing === true;
      }
      return (
        // Assert no AJAX request is running (via jQuery or Drupal) and no
        // animation is running.
        (typeof jQuery === 'undefined' || (jQuery.active === 0 && jQuery(':animated').length === 0)) &&
        (typeof Drupal === 'undefined' || typeof Drupal.ajax === 'undefined' || !Drupal.ajax.instances.some(isAjaxing))
      );
    }());
JS;
    $result = $this->getSession()->wait(5000, $condition);
    if (!$result) {
      throw new \RuntimeException('Unable to complete AJAX request.');
    }
  }

```

This is working better, but also not in all cases. I still need to wait manually when e.g. submitting an entity browser widget.

Also not sure if this is 8.x specific.

Berdir:
The merge request was merged, closing this.

jochemvn:
I saw this issue made it in 3.3.1. However now our builds are failing on 'iWaitForAjaxToFinish'. The error is we get is:
```
And I wait for AJAX to finish                                                                                 # SocialMinkContext::iWaitForAjaxToFinish()
      Unable to complete AJAX request. (RuntimeException)
```
Our usecase is that we test uploading a file in a normal image field and we wait for the Ajax process to finish the upload and then we press save. 

Any thoughts on how to fix this? For now I've changed the version back to 3.3.0.

Berdir:
And how long does that take? The exception is thrown if the condition is not true after 5s. Before it just waited and then moved on. We *could* go back to that, but it is likely that this didn't really work either before already, you just didn't get an error

alexdesignworks:
[...]

PR #401, opened 2017-08-04
Use wait on ajax logic from Drupal 8

Berdir:
See #400


jhedstrom:
This looks good to me, but I'm missing how this is specific to Drupal 8?

Berdir:
I don't know, just meant that this is copied from D8 and I didn't test if the ajax API works the same way in 7.x

jhedstrom:
Ah, I misread the title :)

Issue #412, opened 2017-09-18
3.3.1 #401 breaks ajax-wait for Drupal 7


BramDriesen:
When using 3.3.1 and using the Ajax-wait function, it's no longer working.

`And I wait for AJAX to finish                                                         # Drupal\DrupalExtension\Context\MinkContext::iWaitForAjaxToFinish()
      unknown error: Cannot read property 'some' of undefined
        (Session info: chrome=61.0.3163.91)
        (Driver info: chromedriver=2.32.498550 (9dec58e66c31bcc53a9ce3c7226f0c1c5810906a),platform=Windows NT 6.1.7601 SP1 x86_64) (WebDriver\Exception\UnknownError)`

When locking the composer.json to 3.3.0 it's working fine. Please either roll back the change, or fix the issue. I have no idea what's wrong with it.

Berdir:
There is already https://github.com/das-peter/drupalextension/pull/1 but apparently that was created in the wrong Repo, can you test that?

alexdesignworks:
I can confirm that this is an issue.

@Berdir can we get that PR applied here?

das-peter:
@Berdir Thanks for the ping. Created proper pull request here: #414 
Sorry for being a moron and delaying this with my github incompetence ;)

PR #414, opened 2017-09-28
Follow-up #400: iWaitForAjaxToFinish() broken for D7

das-peter:
The issue fixed in #400 breaks the D7 check.
Only thing missing is a check if ` Drupal.ajax.instances` actually exists - which is only the case in D8.
Without the check we get an error like "Call method some on undefined".
While checking the code in D7 I realized we've instances with the ajaxing flag there as well - we just need to access the instances in a different way.
I've tried to do that in this change too.

So far only tested with a D7 installation.


BramDriesen:
This is related to https://github.com/jhedstrom/drupalextension/issues/412

jhedstrom:
This needs to be rebased off of the latest `master` to resolve the test failures.

jhedstrom:
I've merged the latest master in #429.

Issue #413, opened 2017-09-26
Drupal 8.4


nicrodgers:
We've updated our site to 8.4-rc2 and wanted to run our behat features against it, to see if there are any errors introduced.

However, after updating behat and the behat drupal extension, when we try to run behat we now get the following error:
```
In CheckDefinitionValidityPass.php line 61:
                                                                                                                                                                                                                                                                                                              
  The definition for "logger.channel.default" has no class. If you intend to inject this service dynamically at runtime, please mark it as synthetic=true. If this is an abstract definition solely used by child definitions, please add abstract=true, otherwise specify a class to get rid of this error.  
                                                                                  
```

nicrodgers:
Our composer.json for the behat test, incase that helps:
```
{
  "minimum-stability": "dev",
  "require": {
    "drupal/drupal-extension": "dev-master",
    "guzzlehttp/guzzle" : "^6.0@dev",
    "drupal/drupal-driver": "dev-master"
},
  "config": {
    "bin-dir": "bin/"
  }
}
```

pfrenssen:
Hi, thanks for the report. Can you use the verbose option to generate a backtrace to see if this error is actually caused by drupalextension and not by one of your other dependencies?

The `CheckDefinitionValidityPass` class is not part of drupalextension, it seems to belong to the Symfony Dependency Injection component. We don't have any references in our code base to a service named `logger.channel.default` so it is possible this error originates from another project.

nicrodgers:
Hi, I've traced it down a bit further. We get the error whenever our scenarios try to use the step "Given the cache has been cleared".

Here's the backtrace if that's any help:
```
In CheckDefinitionValidityPass.php line 61:
                                                                                                                                                                                                                                                                                                              
  [Symfony\Component\DependencyInjection\Exception\RuntimeException]                                                                                                                                                                                                                                          
  The definition for "logger.channel.default" has no class. If you intend to inject this service dynamically at runtime, please mark it as synthetic=true. If this is an abstract definition solely used by child definitions, please add abstract=true, otherwise specify a class to get rid of this error.  
                                                                                                                                                                                                                                                                                                              

Exception trace:
 Symfony\Component\DependencyInjection\Compiler\CheckDefinitionValidityPass->process() at /Volumes/www/gitlab/govwalesd8/tests/behat/vendor/symfony/dependency-injection/Compiler/Compiler.php:143
[...]

PR #418, opened 2017-11-15
Update core/remove special composer handling when testing

### LINKED ISSUE #417: Remove special composer handling for testing
There's a (forgotten by most) special `composer.json` file currently used for D8 testing. We should see if we can remove that.

We can also update to test 8.4 at the same time. I didn't realize 8.1 was hard-coded.

jhedstrom:
Fixes #417.


pfrenssen:
The failure appears to be related to #413

jhedstrom:
Indeed. I'm taking the opportunity to use composer to install Drupal properly for 8 and beyond, as that will resolve the issue, and ensure against future conflicts (hopefully).

jhedstrom:
Actually, it is still non-trivial to install Drupal as a test dependency. The on-failing version of dependency injection isn't even picked up since the lock file of core isn't processed. For now I'm pinning the version of that library during testing.

PR #393, opened 2017-07-14
[Reroll] WIP: Create an AuthenticationManager service

jonathanjfshaw:
Continues #339, small reroll to address collision with #385

PR #339, opened 2017-01-04
WIP: Create an AuthenticationManager service

pfrenssen:
This is part of #316 which is an initiative to move all code in `RawDrupalContext` to services so its logic can be changed by projects that have customized the way they interact with Drupal.

In this PR an `AuthenticationManager` service will be provided that takes care of authenticating users in Drupal. We will provide a default implementation that matches the UI that Drupal core delivers out of the box.

People that need to customize the authentication procedure can swap out the default implementation with their own. Example use cases: Email Registration module, OAuth, Single Sign On.

PR #425, opened 2017-12-01
Create an AuthenticationManager service

jhedstrom:
Continuation of #339 and #393. This does not contain the logout hardening commit from #393 as I think that can be a separate issue.


pfrenssen:
Oh wow, thanks for picking this back up!

Issue #62, opened 2014-09-14
Improve performance of logout


Berdir:
No pull requests yet, I'm experimenting with some changes in overridden methods but not all of that can be assumed to be generally true.

Here's what I'm currently doing/thinking about:
1. For logout, instead of doing a real logout, I just do $this->getSession()->reset(), which drops any session cookies.
2. For loggedIn(), I'm not doing the initial request to the frontpage and instead just assume that the page I'm on has the logout link. This has a few problems, notably that some of the assert logged in as user with role and so on methods initially call loggedIn() to check if already logged in with a user that matches the condition. In this case, I'm currently just catching the thrown exception and consider it to be not logged in.
3. However, I think those checks are not necessary anyway. I don't know how others structure their tests, but having multiple identical "Given I am logged in as a user with role "X" seems very unlikely and currently (without my hack from above), the overhead of doing an additional frontpage request is way bigger than a few saved user creations here and there.
4. A problem I had with my pretty big test suite (currently 185 scenarios, 2.2k steps, doing sometimes pretty crazy stuff) was that I had some "Given I am an anyomous user" checks on some admin pages, which had no Log Out link. However, similar to 3., with the fast "logout" from 1., just calling logout() should be faster than loading the frontpage to avoid calling it if we're not logged in.

Thoughts?

I'm currently doing 1, 2 and 4 (didn't bother updating all the methods for 3, just catching the exception for now) and that saves me 2 requests for the first login and three for any following (due to the logout) in the same scenario. That sums up quite nicely.

pfrenssen:
1. Great idea. This will be really fast.
2. Both D7 and D8 set a class on the body element that can be used to check if the user is logged in. It should be present on all pages, but this is not guaranteed on custom themes. Then again, the presence of the "Log out" link it also not guaranteed :) We can also check if the session cookie is present and matches a logged in user in the database.
3. This is indeed not common. If you do need to relog in a test, you can work around this by simply adding a step to log out manually before logging in again.
4. Sounds good, requests are slow, we should only do the minimum required.

ybabel:
+1

jonathanjfshaw:
Given @pfrenssen said in #398:  

> I think this whole logic should be greatly simplified and should only cover the login functionality of the base themes that are shipped with Drupal core. Themes can override this completely so there is no way for us to cover all possibilities. With the AuthenticationManager service this will be trivial to override.

I think the way forward here is to concentrate on implementing 1:

[...]

PR #427, opened 2017-12-01
Logout performance

jhedstrom:
This addresses #62 and includes the logout hardening commit from #393.


jhedstrom:
To take full advantage of this, the `logout` method in `RawDrupalContext` should probably take a `$fast` parameter (and the method here should be added to a separate interface). Then steps that don't require an actual logout (eg, `assertAnonymousUser`, etc) can pass in `TRUE` to utilize the fast logout. This can also be used during user cleanup.

jhedstrom:
I've decided to merge this and tag an alpha release of 4.0. We can do follow-ups if folks find issues.

PR #445, opened 2017-12-12
Reroll: Big pipe compatibility follow-up

jhedstrom:
Merged latest `master` for #325.


jhedstrom:
I'm re-opening the original since that's where the current discussion is, and this isn't as simple as just merging in master :)

Issue #446, opened 2017-12-13
Fatal error when running test suite on most recent master


m4olivei:
Since updating to the latest master our Behat test suite is consistently failing with this message:

Behat\Testwork\Call\Exception\FatalThrowableError: Fatal error: Call to a member function deleteAllCookies() on null in /var/www/html/vendor/behat/mink-selenium2-driver/src/Selenium2Driver.php:359

The only relevant change that I can see is that the drupal/drupal-extension package was updated from `ca50162dada22330953f09822f1331bf3aec89b3` to `2eb9232390cf9c251884da713c761131635692ba`.

m4olivei:
Pinning to `ca50162dada22330953f09822f1331bf3aec89b3` fixed the issue. Here is the whole stack trace of the error:

```
Given I am logged in as a user with the "administrator" role               # Drupal\DrupalExtension\Context\DrupalContext::assertAuthenticatedByRole()
      Behat\Testwork\Call\Exception\FatalThrowableError: Fatal error: Call to a member function deleteAllCookies() on null in /var/www/html/vendor/behat/mink-selenium2-driver/src/Selenium2Driver.php:359
      Stack trace:
      #0 /var/www/html/vendor/behat/mink/src/Session.php(103): Behat\Mink\Driver\Selenium2Driver->reset()
      #1 /var/www/html/vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php(133): Behat\Mink\Session->reset()
      #2 /var/www/html/vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php(48): Drupal\DrupalExtension\Manager\DrupalAuthenticationManager->fastLogout()
      #3 /var/www/html/vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/RawDrupalContext.php(530): Drupal\DrupalExtension\Manager\DrupalAuthenticationManager->logIn(Object(stdClass))
      #4 /var/www/html/vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/DrupalContext.php(66): Drupal\DrupalExtension\Context\RawDrupalContext->login(Object(stdClass))
      #5 [internal function]: Drupal\DrupalExtension\Context\DrupalContext->assertAuthenticatedByRole('administrator')
[...]

PR #447, opened 2017-12-13
#446 Check for session first before logging out in fastLogout method of DrupalAuthenticationManager

m4olivei:
Attempt at resolving #446.

### DIFF
--- a/src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
+++ b/src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
@@ -133,7 +133,7 @@ class DrupalAuthenticationManager implements DrupalAuthenticationManagerInterfac
         $session = $this->getSession();
         if ($session->isStarted()) {
             $session->reset();
-            $this->userManager->setCurrentUser(false);
         }
+        $this->userManager->setCurrentUser(false);
     }
 }
--- a/src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
+++ b/src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
@@ -132,8 +132,8 @@ class DrupalAuthenticationManager implements DrupalAuthenticationManagerInterfac
     {
         $session = $this->getSession();
         if ($session->isStarted()) {
-          $session->reset();
-          $this->userManager->setCurrentUser(FALSE);
+            $session->reset();
+            $this->userManager->setCurrentUser(false);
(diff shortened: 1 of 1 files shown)


m4olivei:
This patch works for me. I think it's a Selenium specific issue, where the Selenium Mink Driver is assuming that the session is already started when you call `reset` here: https://github.com/minkphp/MinkSelenium2Driver/blob/v1.3.1/src/Selenium2Driver.php#L359. When it's not, 💥 .

m4olivei:
Incidentally, I see there is some awesome test coverage in this package. I wonder if it would be possible to have some test cases with Mink Selenium2 Driver and a browser like Chrome?

jhedstrom:
> Incidentally, I see there is some awesome test coverage in this package. I wonder if it would be possible to have some test cases with Mink Selenium2 Driver and a browser like Chrome?

That would be great! Happy to review PRs that update `.travis.yml` to add some coverage there (and perhaps a new suite or profile in `behat.yml.dist` to specifically target selenium...

src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
```diff
@@ -130,7 +130,10 @@ public function loggedIn()
      */
     public function fastLogout()
     {
-        $this->getSession()->reset();
-        $this->userManager->setCurrentUser(false);
+        $session = $this->getSession();
+        if ($session->isStarted()) {
+            $session->reset();
+            $this->userManager->setCurrentUser(false);
```
jhedstrom: I wonder if we should unset the user outside the conditional?

m4olivei: That makes sense, will update.

m4olivei:
> That would be great! Happy to review PRs that update .travis.yml to add some coverage there (and perhaps a new suite or profile in behat.yml.dist to specifically target selenium...

I took an honest look at it (being new to Travis), and it seems like it would be tricky to get all the dependencies right to be able to run chrome in a window manager and hook selenium up to it. I've seen it done in CircleCI before. Although it would be nice, I'm going to slowly climb out of this rabbit hole for now :).

jhedstrom:
Thanks!

jhedstrom:
@m4olivei I found [this Travis setup for the mink selenium driver](https://github.com/minkphp/MinkSelenium2Driver/blob/master/.travis.yml), it might help you out if you get the chance :)

PR #449, opened 2017-12-17
drupalExtension is fully translated to Spanish

jose-carmona:
Steps translated:
* the/these (following )languages are available:
* users:
* I should see (the text ):text in the :rowText row
* I should not see (the text ):text in the :rowText row
* I am logged in as a/an :role
* I (should ) see the button :button
* I (should ) see the :button button
* I should not see the button :button
* I should not see the :button button
* I should see the button :button in the :region( region)
* I should see the :button button in the :region( region)
* I( should) see the :tag element in the :region( region)
* I( should) not see the :tag element in the :region( region)
* I( should) not see :text in the :tag element in the :region( region)
* I( should) see the :tag element with the :attribute attribute set to :value in the :region( region)
* I( should) see :text in the :tag element with the :attribute attribute set to :value in the :region( region)
* I( should) see :text in the :tag element with the :property CSS property set to :value in the :region( region)

I think with this PR drupalExtension is fully translated to Spanish.

This PR creates an environment for testing translations placing a 'i18n' directory in which to place the translations of the features.

### DIFF
--- a/features/i18n/es/blackbox.feature
+++ b/features/i18n/es/blackbox.feature
@@ -1,0 +1,106 @@
+# language: es
+@blackbox
+Característica: Test DrupalContext
+  # Esta característica es copia traducida de la "feature" correspondiente
+  # para demostrar que los correspondientes pasos están bien traducidos
+  # al español
+  Para probar el adecuado funcionamiento del contexto Drupal usando el driver "blackbox"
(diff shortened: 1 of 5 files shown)


jhedstrom:
This is fantastic! Thanks for the tests too!

PR #459, opened 2018-02-03
Created TestModuleContext.

phenaproxima:
Requires jhedstrom/DrupalDriver#164.

This adds a new context, TestModuleContext, which allows scenarios to automatically enable a module or set of modules before they start, then automatically uninstall those modules after the scenario is completed. This brings the paradigm of test modules, used extensively in Drupal core, to Behat.

To use it, a scenario (or feature) can add tags of the format ```@with-module:MODULENAME```, and the module MODULENAME will be installed before the scenario, and uninstalled afterwards.

This is already implemented (and works!) over in acquia/lightning-dev.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -8,6 +8,7 @@ default:
         - Drupal\DrupalExtension\Context\MessageContext
         - Drupal\DrupalExtension\Context\MinkContext
         - Drupal\DrupalExtension\Context\MarkupContext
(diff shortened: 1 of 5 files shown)


jhedstrom:
The tests are failing due to some coding standard issues, and also passing before scenario contexts to after scenario hooks (or the other way around). Should be a pretty quick fix I think.

PR #460, opened 2018-02-06
replace links to drush.ws with drush.org --drush.ws contains unrelated clickbait content

ownsourcing:


jhedstrom:
Great! Thanks for this!

PR #462, opened 2018-02-08
Lock DrupalDriver to the latest stable release

### LINKED ISSUE #461: Lock the version of DrupalDriver
We are closely tied to DrupalDriver, and it often happens that for a certain improvement to be made in drupalextension we need a sister issue to provide the relevant functionality in DrupalDriver first. This is probably the reason that in `composer.json` we are following the latest master branch of DrupalDriver:

```
  "require": {
    ...
    ...
  },
```

The drawback of this is that the users of drupalextension are always receiving the very latest development commits of DrupalDriver and this can cause tests to unexpectedly start failing after doing a composer update.

Let's lock on the latest release of DrupalDriver. If new features are in development that rely on bleeding edge commits in DrupalDriver we can always temporarily unlock it while working on the PR. Then when the work is well tested and ready for merging in develop we can sync it with a DrupalDriver release, and update the version in the composer.json file.

pfrenssen:
Fixes #461

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     "behat/mink-extension": "~2.0",
     "behat/mink-goutte-driver": "~1.0",
     "behat/mink-selenium2-driver": "~1.1",
-    "drupal/drupal-driver": "dev-master",
+    "drupal/drupal-driver": "^1.3.2",
     "symfony/dependency-injection": "~2.7|~3.0",
     "symfony/event-dispatcher": "~2.7|~3.0"
   },


jhedstrom:
I think this makes sense.

jonathanjfshaw:
Composer is telling me that this makes drupal-extension incompatible with drupal-driver dev-master; in other words dev-master does not satisfy the constraint ^1.3.2. This makes it hard (impossible?) for users of the drupal extension to opt-in to the bleeding edge of the driver.

PR #463, opened 2018-02-09
remove sudo from composer command

ownsourcing:
Composer maintainers strongly advise against running composer as sudo/root. On Mac and Ubuntu it appears not to be necessary. Is it possible to accomplish everything without using sudo with composer?

See their FAQ:
How do I install untrusted packages safely? Is it safe to run Composer as superuser or root?
https://getcomposer.org/doc/faqs/how-to-install-untrusted-packages-safely.md

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ matrix:
       env: DRUPAL_VERSION=6
   allow_failures:
     - php: 7.0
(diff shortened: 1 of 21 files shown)


ownsourcing:
arg. edited out of date file. easier just to close this one and open a new one.

PR #467, opened 2018-02-23
Allows AJAX timeout to be overridden.

### LINKED ISSUE #466: Allow the `iWaitForAjax` timeout to be configurable
When sites are under heavy load, certain AJAX calls can take longer than the 5 seconds we have coded into `MinkContext::iWaitForAjax()`:

```php
        $result = $this->getSession()->wait(5000, $condition);
```

Making this configurable via a `behat.yml` parameter would allow folks to override this as needed.

jhedstrom:
- Fixes #466

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -11,7 +11,7 @@ default:
       filters:
         tags: "@blackbox"
   extensions:
-    Behat\MinkExtension:
+    Drupal\MinkExtension:
       goutte: ~
       zombie: ~
       base_url: http://127.0.0.1:8888/blackbox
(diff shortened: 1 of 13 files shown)


jhedstrom:
Adding @pfrenssen as a reviewer if you have time. This is a simple feature, but required a substantial change in terms of the preferred way of specifying how the mink extension is configured.

behat.yml.dist
```diff
@@ -11,7 +11,7 @@ default:
       filters:
         tags: "@blackbox"
   extensions:
-    Behat\MinkExtension:
+    Drupal\MinkExtension:
```
davereid: Could this change use more documentation in the release notes?

jhedstrom: Good call! I've update https://github.com/jhedstrom/drupalextension/releases/tag/v4.0.0alpha3

Issue #479, opened 2018-03-20
Provide more verbose output when iWaitForAjaxToFinish() fails.


marcoscano:
Currently error messages when an AJAX call fails are quite short:

```
AfterStep # Drupal\DrupalExtension\Context\MinkContext::afterJavascriptStep()
      Unable to complete AJAX request. (RuntimeException)
```

or 

```
And I wait for AJAX to finish                                                                                 # MinkContext::iWaitForAjaxToFinish()
      Unable to complete AJAX request. (RuntimeException)
```

Could we add to that exception message more information about the event that failed?

brooke-heaton:
`iWaitForAjaxToFinish` appears to be inherently problematic function. All tests are now failing at this step for me on 4.1 using `selenium/standalone-chrome:3.141.59-oxygen`

mxr576:
Related #557

PR #480, opened 2018-03-20
#479: Provide more verbose exception message when AJAX fails.

marcoscano:
Very rough, just as a proof-of-concept, but this seems to help a lot while troubleshooting where a given error is coming from.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MinkContext.php
+++ b/src/Drupal/DrupalExtension/Context/MinkContext.php
@@ -127,7 +127,7 @@ class MinkContext extends MinkExtension implements TranslatableContext
    *
    * @Given I wait for AJAX to finish
    */
-    public function iWaitForAjaxToFinish($event = NULL)
+    public function iWaitForAjaxToFinish($event = null)
     {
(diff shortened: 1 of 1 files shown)


src/Drupal/DrupalExtension/Context/MinkContext.php
```diff
@@ -98,7 +98,7 @@ public function beforeJavascriptStep($event)
         }
         $text = $event->getStep()->getText();
         if (preg_match('/(follow|press|click|submit)/i', $text)) {
-            $this->iWaitForAjaxToFinish();
+            $this->iWaitForAjaxToFinish($event);
```
jhedstrom: This variable doesn't appear to be defined anywhere...

Berdir: `public function beforeJavascriptStep($event)`

It is passed to the method :)

jhedstrom: haha, oops.

jhedstrom:
Great, thanks!

PR #484, opened 2018-04-16
Adds a trait for determining current scenario tags.

### LINKED ISSUE #468: The `beforeJavascriptStep` doesn't check all tags
This code in `MinkContext::beforeJavascriptStep()` doesn't find tags at the scenario level, only at the feature-file level:

```php
    public function beforeJavascriptStep($event)
    {
        /** @var \Behat\Behat\Hook\Scope\BeforeStepScope $event */
        $tags = $event->getFeature()->getTags();
        if (!in_array('javascript', $tags)) {
            return;
        }
```
instead, it should check both the feature, and the scenarios, something like:

```php
  $tags = $event->getFeature()->getTags();
  foreach ($event->getFeature()->getScenarios() as $scenario) {
    $tags = array_merge($tags, $scenario->getTags());
  }

jhedstrom:
- Fixes #468 

I found a proper workaround over in https://github.com/Behat/Behat/issues/703#issuecomment-86687563, and have added that here as a trait.


pfrenssen:
Very useful, thanks!

PR #489, opened 2018-05-17
Authenticate user in the backend bootstrap process on login.

### LINKED ISSUE #488: Log user in on the backend during login steps
Many modules, and services utilize the `current_user` service for determining ownership, etc. When these hooks or services are called from the backend (for instance, during `Given I am viewing my node...`), `current_user` will be the anonymous user on the backend.

It would be good to have the login and logout steps log the user in on the backend (for the Drupal driver).

jhedstrom:
- Fixes #488 

This could use some testing, especially from folks using alternative authentication managers.

The corresponding driver PR is at https://github.com/jhedstrom/DrupalDriver/pull/187

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     "behat/mink-extension": "~2.0",
     "behat/mink-goutte-driver": "~1.0",
     "behat/mink-selenium2-driver": "~1.1",
-    "drupal/drupal-driver": "^2.0.0",
+    "drupal/drupal-driver": "dev-186-backend-auth",
     "symfony/dependency-injection": "~3.0",
     "symfony/event-dispatcher": "~3.0"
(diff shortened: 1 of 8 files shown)


jhedstrom:
@pfrenssen any word on how this worked with your test suites?

composer.json
```diff
@@ -27,7 +27,7 @@
     "behat/mink-extension": "~2.0",
     "behat/mink-goutte-driver": "~1.0",
     "behat/mink-selenium2-driver": "~1.1",
-    "drupal/drupal-driver": "^2.0.0",
+    "drupal/drupal-driver": "dev-186-backend-auth",
```
jhedstrom: Note this needs to be undone before merging.

Issue #498, opened 2018-08-14
ConfigContext config set cannot be run multiple times in one scenario


upchuk:
The `ConfigContext::setConfig()` keeps a backup of the config that we set. However, if we need to set it multiple times in the same scenario the backup is overridden by the last time it runs. The fix is pretty simple: check if the backup has already been made already before setting it.

PR will follow.

pfrenssen:
This looks good to me. Thanks for the contribution!

Looking at this code I am thinking that the `setConfig()` method should be made protected. The public version of this is called `setBasicConfig()` and has a step definition associated with it. I will make a followup for this.

PR #499, opened 2018-08-14
GH-498 - Fixing ConfigContext backup strategy.

upchuk:
Fixes GH-498.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/ConfigContext.php
+++ b/src/Drupal/DrupalExtension/Context/ConfigContext.php
@@ -107,6 +107,13 @@ class ConfigContext extends RawDrupalContext implements TranslatableContext
     {
         $backup = $this->getDriver()->configGet($name, $key);
         $this->getDriver()->configSet($name, $key, $value);
-        $this->config[$name][$key] = $backup;
+        if (!array_key_exists($name, $this->config)) {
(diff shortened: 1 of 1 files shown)


pfrenssen:
This looks good to me. Thanks for the contribution!

src/Drupal/DrupalExtension/Context/ConfigContext.php
```diff
@@ -107,6 +107,13 @@ public function setConfig($name, $key, $value)
     {
         $backup = $this->getDriver()->configGet($name, $key);
         $this->getDriver()->configSet($name, $key, $value);
-        $this->config[$name][$key] = $backup;
+        if (!array_key_exists($name, $this->config)) {
+            $this->config[$name][$key] = $backup;
+            return;
+        }
+
+        if (!array_key_exists($key, $this->config[$name])) {
```
pfrenssen: As a very small optimization we can rewrite both if statements as one `isset()`:

```
if (!isset($this->config[$name][$key])) {}
```

pfrenssen:
```
php > $array['key'] = NULL;
php > var_dump(isset($array['key']));
bool(false)
```

;)

PR #502, opened 2018-08-20
RawDrupalContext::loggedIn() can return false positive

trwill:
The default `login_form_selector` is using `form#user-login` as a selector, but the actual drupal login form (at least for current d8) uses `form#user-login-form` as the id. My guess is an update to underlying form changed this at some point (maybe during migration to drupal 8?)? 

This can ultimately cause false positives from `RawDrupalContext::loggedIn()` - in some scenarios the login is validated by manually checking the  `/user/login` page and looking for the form via this selector - if its not found, the user is assumed to be logged in. In this case even though the user is anonymous, the form is not found, so it returns true. 

I left the old selector in for any possible BC issues, similar to how the `logged_in_selector` works. 

For example, here's a blank site spun up w/ simplytest.me showing the actual selector:
image

### DIFF
--- a/src/Drupal/DrupalExtension/ServiceContainer/DrupalExtension.php
+++ b/src/Drupal/DrupalExtension/ServiceContainer/DrupalExtension.php
@@ -145,7 +145,7 @@ class DrupalExtension implements ExtensionInterface
             scalarNode('success_message_selector')->end()->
             scalarNode('warning_message_selector')->end()->
             scalarNode('login_form_selector')->
-              defaultValue('form#user-login')->
+              defaultValue('form#user-login,form#user-login-form')->
             end()->
             scalarNode('logged_in_selector')->
               defaultValue('body.logged-in,body.user-logged-in')->


pfrenssen:
This depends entirely on which theme you are using. The exact selector to use can be configured in `behat.yml` and it is possible to override the logic by swapping out the AuthenticationManager service.

trwill:
I understand it can be overridden but this is from default theme - shouldn't we be setting the default?

jhedstrom:
I can't figure how it's working with the old selector as the form ID doesn't appear to have changed in Drupal 8 since 2013...

trwill:
@jhedstrom yeah it certainly appears to have been introduced with the launch of Drupal 8. It only 'falls back' to this logic of checking for the form if the `<body>` el [does not contain the 'logged_in_selector`](https://github.com/jhedstrom/drupalextension/blob/master/src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php#L109). I realize this may be a bit of an edge case, but just needs a simple update for Drupal 8 defaults and will go back to working as intended.

jhedstrom:
This fix makes sense to me. Thanks!

Issue #501, opened 2018-08-17
ConfigContext needs to use the original config when backing up, not the overridden one


upchuk:
Inside the `ConfigContext::setConfig()` we need to use the original (DB value) config instead of the potentially overridden value. For this, we need a new method on the Driver interface. See related issue on the Driver for this: https://github.com/jhedstrom/DrupalDriver/issues/194

PR will follow.

PR #503, opened 2018-08-28
Use new configGetOriginal method from DrupalDriver.

aritomelo:
The ConfigContext needs to use the "configGetOriginal()" method when backing up the config value.

**Issues also created on Github:**
- https://github.com/jhedstrom/drupalextension/issues/501

- https://github.com/jhedstrom/DrupalDriver/issues/194


aritomelo:
The build failed once the new method it is not yet merged on DrupalDriver component, so it does not exist!

PR #510, opened 2018-10-01
Provide TagTrait to replace ScenarioTagTrait.

### LINKED ISSUE #509: `ScenarioTagTrait` is not intuitive to use
I have a step definition that depends on the `@email` tag being present and I tried retrieving this using the `ScenarioTagTrait`. It appears this trait will only retrieve the tags for both the scenario and the feature inside a `@beforeStep` hook, it won't work in any other hooks or step definitions. `::getCurrentScenarioTags()` expects a `StepScope` to be passed, this is only available in `@beforeStep` hooks. It looks this is the case because this has been written specifically for `MinkContext::beforeJavascriptStep()`. It seems it should be possible to make this more universally usable by making a few modifications.

* It would be handy to always return the tags for the scenarios as well as the features. To make it clear this works not only for scenario tags I would rename this from `ScenarioTagTrait` to just `TagTrait`.
* In addition to `::registerScenario()` we can implement a `::registerFeature()` method that keeps track of the current feature, and a `::getFeature()` method to return it.
* `::getCurrentScenarioTags()` could be renamed to `::getTags()` and doesn't need a `StepScope` passed into it any more. It can instead call `::getFeature()`.
* Since most of the time the developer would be interested whether a particular tag is present, it would be a nice addition to add a `::hasTag(string $tag): bool` method.

pfrenssen:
Fixes #509.


pfrenssen:
I have split up the original `ScenarioTagTrait` into three separate traits since all three are useful on their own:

* `FeatureTrait`: allows to access the current `FeatureNode` inside a Context class.
* `ScenarioTrait`: allows to access the current `ScenarioNode` inside a Context class.
* `TagTrait`: allows to retrieve the current tags from the feature or scenario.

The `TagTrait` depends on the two other traits, but instead of forcing a developer to include all three traits, I wrapped the two depending traits in a helper class which is included as a singleton through a getter method. This makes it easy to use the trait by just adding `use TagTrait;` to the context class.

pfrenssen:
This is failing on PHP 5.5 since the PR is using argument unpacking which was introduced in PHP 5.6.

@jhedstrom are you planning on still supporting PHP 5.5 on the 4.x branch?

jhedstrom:
@pfrenssen I'd prefer not to. Honestly, I'd love to only support 7.x, but don't know if that will work for folks still running Drupal 6 or not.

pfrenssen:
Maybe we should do a poll on Twitter to see if people are still interested in Drupal 6 support for BDE 4.x?

We could also maybe decide to keep supporting 3.x for D6 only, say for one or two years more? We could keep all D7 and D8 issues for 4.x.

I think this will not be a lot of maintenance work, I haven't seen many D6 issues in the queue lately :)

I personally would love to move to PHP 7 for the 4.x branch. But I think we should wait a few months, PHP 5.6 is still officially supported until December of this year.

pfrenssen:
Added support for PHP 5.5, we can have the discussion about which PHP and Drupal versions to support in the future in a dedicated issue :)

jhedstrom:
Makes sense to me! Thanks!

PR #511, opened 2018-10-02
Start supporting Drupal 7 on PHP 7.1

pfrenssen:
When PHP 7.1 was initially released there were some incompatibilities for Drupal 7 and we temporarily allowed this to fail on Travis. In the meantime D7 works fine on PHP 7.1 and we have a green build!

Let's support this combo from now on.


jhedstrom:
That's great!

PR #512, opened 2018-10-03
Start testing on PHP 7.2

pfrenssen:
PHP 7.2 has been stable for some time now, let's add it to the test matrix and find out how well we support it!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.6
   - 7.0
   - 7.1
+  - 7.2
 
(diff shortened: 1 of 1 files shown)


pfrenssen:
The test on Drupal 7 is failing because of [Issue #2931464: Features PHP 7.2 Function create_function() is deprecated](https://www.drupal.org/project/features/issues/2931464). This has been 

Since this is a bug in the Features module and not in our code we could perhaps opt to apply the patch to Features.

PR #514, opened 2018-10-11
Add a note about the need to remove the entries in behat.yml to use BEHAT_PARAMS

jonpugh:


jhedstrom:
Yeah, it's crazy that Behat takes values out of the yml over environment variables. Tests failed for unrelated reasons, so I'm committing this. Thanks!

PR #516, opened 2018-10-15
Warn users when the message table is not correctly formatted

### LINKED ISSUE #515: MessageContext does not warn about improper use of message tables
Whenever a list of messages is checked in `MessageContext`, the step definitions expect to be passed a table of messages with a header indicating the message type, such as `error messages`:

```
Then I should see the following warning messages:
  | warning messages                                                                     |
  | You are viewing the published version. To view the latest draft version, click here. |
```

When this header is not present the messages will not be checked. Currently no warning is being emitted when this happens, so when a developer uses the following table the test will be silently skipped:

```
Then I should see the following warning messages:
  | You are viewing the published version. To view the latest draft version, click here. |
```

It is likely that developers miss this sometimes in their tests since it is not documented that the header is required, and is only apparent when reading the code. We recently discovered this and did a check in our existing test suite and found that in numerous occasions the header was missed: https://github.com/ec-europa/joinup-dev/pull/1388

It would be nice if we could improve the documentation to clearly indicate that this header is required, and also fail the test if the header is missing.

pfrenssen:
Fixes #515

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MessageContext.php
+++ b/src/Drupal/DrupalExtension/Context/MessageContext.php
@@ -41,14 +41,17 @@ class MessageContext extends RawDrupalContext implements TranslatableContext
   /**
    * Checks if the current page contains the given set of error messages
    *
-   * @param $messages
-   *   array An array of texts to be checked
+   * @param array $messages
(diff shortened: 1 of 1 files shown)


jhedstrom:
This looks good to me!

PR #521, opened 2018-12-18
Corresponding list handler fixes for the Drupal Driver.

jhedstrom:
- Changes to test compatiblity with https://github.com/jhedstrom/DrupalDriver/issues/168

It's odd these scenarios were previously passing--the actual field value being saved on the node was `One`, rather than the field key of `1` and `2`.


jhedstrom:
Note the composer change needs to be reverted prior to merge.

PR #523, opened 2019-01-08
Adding the option to specify a custom login/logout path in your behat.yml

digitole:
Inspired by https://www.drupal.org/project/drupalextension/issues/2958158

In certain circumstances it can be a good idea to change the standard drupal login path (ie: for security improvements).
In that case, the Drupal Behat extension will not be able to log in to the standard /user URL.

I've developed a little patch that adds a further configuration option to specify custom login/logout URL.

### DIFF
--- a/doc/blackbox.rst
+++ b/doc/blackbox.rst
@@ -89,6 +89,8 @@ customized the label visible to users, you can change that text as follows:
 
      Drupal\DrupalExtension:
        text:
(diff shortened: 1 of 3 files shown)


rickkowal:
I like and need this -- is there a chance this will be merged soon?

jhedstrom:
This makes sense to me. Thanks!

PR #525, opened 2019-01-23
Language neutral login.

### LINKED ISSUE #524: Log in when current language different from English
v3.4.1. – I have a French site here. Actually it's just language negotiation set to use Domain instead of path prefixes. And I have an `edit.feature` with a scenario that just ensures certain roles can log in and edit random nodes:

```
Feature: Let's login as users with certain roles and check that node edit
  forms are accessible.

  @api
  Scenario: Login as publisher and try to edit some random node forms
    Given I am logged in as a publisher
    # ...
```

Logging in using the English base URL `english-site.localhost` works just fine. But running the same test against the French base URL `frensh-site.localhost` gives me following error:

>     Given I am logged in as a publisher                               # Drupal\DrupalExtension\Context\DrupalContext::assertAuthenticatedByRole()
>       Form field with id|name|label|value|placeholder "Password" not found. (Behat\Mink\Exception\ElementNotFoundException)

I then might have found the culprit here:

https://github.com/jhedstrom/drupalextension/blob/50ff0f413f0dc4732f49638e743f86f45e835e50/src/Drupal/DrupalExtension/Context/RawDrupalContext.php#L550-L552

When I change the above to use the `name` attributes of the input elements and the translated "Log in" string it starts working again.

    $element->fillField('name', $user->name);
    $element->fillField('pass', $user->pass);
    $submit = $element->find('xpath', '//form[@id="user-login-form"]')->find('xpath', '//input[@type="submit"]');

Do we need to fix this? Or is there already some `getTranslationFromContext` kind of functionality implemented at some place and I only need to define a `default_language` or something like that in my `behat.yml`?

Refs #93

leymannx:
Closes #524

### DIFF
--- a/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
@@ -547,9 +547,9 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface {
 
     $this->getSession()->visit($this->locatePath('/user'));
     $element = $this->getSession()->getPage();
-    $element->fillField($this->getDrupalText('username_field'), $user->name);
-    $element->fillField($this->getDrupalText('password_field'), $user->pass);
-    $submit = $element->findButton($this->getDrupalText('log_in'));
+    $element->fillField('name', $user->name);
+    $element->fillField('pass', $user->pass);
+    $submit = $element->find('xpath', '//form[@id="user-login-form"]')->find('xpath', '//input[@type="submit"]');
     if (empty($submit)) {
       throw new \Exception(sprintf("No submit button at %s", $this->getSession()->getCurrentUrl()));
     }


leymannx:
Tests failing due to some Drush issue:

> 0.10s$ drush --yes --root=$PWD/drupal site-install --db-url=mysql://travis:@127.0.0.1/drupal
> The program 'drush' is currently not installed. To run 'drush' please ask your administrator to install the package 'drush'
> The command "drush --yes --root=$PWD/drupal site-install --db-url=mysql://travis:@127.0.0.1/drupal" failed and exited with 127 during .
> Your build has been stopped.

leymannx:
Ah yo, now I found out this actually would be the start to make https://behat-drupal-extension.readthedocs.io/en/3.1/blackbox.html#override-text-strings needless. Which probably shouldn't happen in v3 anymore. But maybe in v4.

I guess this PR can get closed then. And I'll have a look into the v4 branch. 😃

pfrenssen:
True this would make the `password_field` and `username_field` text overrides obsolete. I think we can deprecate it though so this could possibly still be a candidate for 3.x which is still supported.

We do prefer changes to be made first for 4.x and then backported to 3.x.

PR #527, opened 2019-02-04
Provide a step to check that a button is not in a region.

idimopoulos:
There is already a step in `MarkupContext` to assert that a button exists within a region but not one to assert that one does not exist. Here is a small patch.

### DIFF
--- a/features/blackbox.feature
+++ b/features/blackbox.feature
@@ -43,7 +43,7 @@ Feature: Test DrupalContext
 
   Scenario: Button not in region
     Given I am on the homepage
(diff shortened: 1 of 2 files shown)


pfrenssen:
Nice addition! Could you add a test for this?

pfrenssen:
This should do the trick, but the test looks for the "content" region on the homepage, but this page has no content. Could you maybe change it to instead look into the "right header"?

pfrenssen:
Great, thanks!

We have a positive and negative step definition for most things, but the negative one was missing for buttons, and this is now fixed :+1:

Issue #486, opened 2018-04-30
How to change/set ajax_timeout


miiimooo:
I have found a problem that click/select steps don't wait long enough (or at all?)

Trying to debug it I got this output:
```
│
    ╳  Unable to complete AJAX request. {"name":"step.after","feature":"Terms","step":"I click the text \"Jag vill vara anonym\"","suite":"default","ajax_timeout":null} (RuntimeException)
```
You see I have added some debugging code to display the ajax_timeout. 
The reference in master is here https://github.com/jhedstrom/drupalextension/blob/master/src/Drupal/DrupalExtension/Context/MinkContext.php#L153

I can't work out where to set this. I tried in behat.yml 
```
  extensions:
    Behat\MinkExtension:
      ajax_timeout: 5
```
I got 
```
    In ArrayNode.php line 311:
                                                                                         
  Unrecognized option "ajax_timeout" under "testwork.mink"  
```

jhedstrom:
Since the `ajax_timeout` parameter is provided by this extension, you need to update your extension section to specify `Drupal\MinkExtension` rather than `Behat\MinkExtension`:

```yaml
  extensions:
    Drupal\MinkExtension:
      ajax_timeout: 5
```
Note that `5` is the default behavior, so you may want to try `10` or `15`.

miiimooo:
Thanks @jhedstrom 

Did you notice the error output above? It seems like the default was null there

miiimooo:
I can see a default of 5 in vendor/drupal/drupal-extension/src/Drupal/MinkExtension/ServiceContainer/MinkExtension.php but I don't think that is used in this context

I have vendor/behat/mink-extension/src/Behat/MinkExtension/Context/RawMinkContext.php which provides the public function getMinkParameter($name) which is used by vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/MinkContext.php:iWaitForAjaxToFinish

miiimooo:
Found I had version spaghetti

jhedstrom:
Glad you got it working :)

miiimooo:
I realise this is still happening, independent of behat version.

Could someone confirm that in vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/MinkContext.php:153 they actually get a value for $this->getMinkParameter('ajax_timeout')? 

[...]

PR #528, opened 2019-02-05
Throw a helpful exception if the AJAX timeout is missing

pfrenssen:
In https://github.com/jhedstrom/drupalextension/issues/486#issuecomment-459487280 @leevh mentioned that the AJAX timeout won't work if the user has made a mistake in the configuration and used `Behat\MinkExtension` instead of our overridden `Drupal\MinkExtension`.

This is a very easy mistake to make, especially when considering that our documentation for versions 3.x and lower of DrupalExtension suggests to use `Behat\MinkExtension` instead of `Drupal\MinkExtension`. This is also suggested in a number of blog posts and stack overflow answers. Also people upgrading from 3.x will likely hit this bug.

This can be quite insidious, since everything will seem to work as normal, but when the `@javascript` tag is being used then suddenly mysterious failures will pop up: "Unable to complete AJAX request".

I suggest to check if the timeout is set, and if not let's throw an exception that can help people on the right path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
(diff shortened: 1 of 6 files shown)


jhedstrom:
Looks good!

PR #529, opened 2019-02-15
Update contexts.rst

vivekpan27:


jhedstrom:
This typo has been fixed in `master`, but never backported. Since this is docs-only, I'm committing this even though the branch tests appear to be broken.

PR #558, opened 2019-11-21
Log in with extra fields

marcelovani:
https://www.drupal.org/project/drupalextension/issues/3095934

### DIFF
--- a/src/Drupal/DrupalExtension/Context/DrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/DrupalContext.php
@@ -91,6 +91,7 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext
             // Assign fields to user before creation.
             foreach ($fields->getRowsHash() as $field => $value) {
                   $user->{$field} = $value;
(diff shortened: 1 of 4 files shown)


pfrenssen:
This is unlikely to be accepted since this diverges from the core login form. We cannot support any modification that is done by any possible module or form alter.

In DrupalExtension 4.0 we factored out the `AuthenticationManager` to a separate service  to allow developers to swap it out for their own custom implementation to match the modules and custom code they use. We are keeping our implementation to what ships with Drupal core out of the box.

See https://github.com/jhedstrom/drupalextension/pull/339:

> In this PR an AuthenticationManager service will be provided that takes care of authenticating users in Drupal. We will provide a default implementation that matches the UI that Drupal core delivers out of the box.
> People that need to customize the authentication procedure can swap out the default implementation with their own. Example use cases: Email Registration module, OAuth, Single Sign On.

Here is an example of how to swap out the `AuthenticationManager` with your own custom version using `friends-of-behat/service-container-extension`: https://github.com/pfrenssen/email_registration/pull/1/files

PR #561, opened 2019-12-05
Add space arround the keywords to avoid wrongly matching.

weynhamz:
Trying to fix jhedstrom#560.

Maybe there is a better way.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MinkContext.php
+++ b/src/Drupal/DrupalExtension/Context/MinkContext.php
@@ -103,7 +103,7 @@ class MinkContext extends MinkExtension implements TranslatableContext
             return;
         }
         $text = $event->getStep()->getText();
-        if (preg_match('/(follow|press|click|submit)/i', $text)) {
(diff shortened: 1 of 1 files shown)


pfrenssen:
Thanks! Can you try with word boundaries?

```
'/\b(follow|press|click|submit)\b/i'
```

This is a more robust and "regexy" way to check for individual words in a string.

weynhamz:
Thanks, @pfrenssen, just updated, the boundary check is better, I was just trying to have a quick fix.


Though I still do not think this is a good approach to trigger an AJAX wait, I am too familiar with Behat, is there a better way like using tags or listeners on click event/method to trigger it instead?

pfrenssen:
Thanks!

I don't think it is possible to trigger methods asynchronously in a running PHP process from JS code due to the single threaded nature of PHP, and we are already limiting this on the `@javascript` tag. This is the best we can do that would work with all kinds of Drupal projects.

Of course in your own project you can extend `MinkExtension` and provide any logic that is specific to your project to narrow down the cases where this is triggered.

jhedstrom:
Thanks!

PR #569, opened 2020-02-17
Allow Symfony 4 as dependency

Berdir:
Same as https://github.com/jhedstrom/DrupalDriver/pull/222

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ install:
   # can create conflicts.
   # @todo Change test process to require drupalextension from the local folder
   #   to let composer handle compatibility and simulate a more common process.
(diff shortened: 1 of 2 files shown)


Berdir:
Not sure what exactly the problem is here. Only thing I found was https://github.com/symfony/symfony/issues/29907, could be a conflict between what is used by drupal and what's added by this, because I think we're not actually installing drupalextension inside drupal and then nothing ensures that version are kept in sync.

Berdir:
To make this work and install the proper version of things, we might want to use a similar approach as DrupalCI, and add the git checkout of drupalextension as a repository to drupal an then install it from there, that will then respect existing dependencies and pick a compatible version.

Berdir:
Got it to pass with a workaround and created #570 with ideas on how to improve travis testing.

PR #581, opened 2020-11-24
Fixes #580: correct syntax error in step definition for logging in as…

### LINKED ISSUE #580: Step annotation syntax error: Given I am logged in as a user with the :role role(s) and I have the following fields
I was having trouble getting a step of the form `Given I am logged in as a user with the :role role and I have the following fields:` to run, and Behat would helpfully offer to create a snippet for me, even though there is already a `function assertAuthenticatedByRoleWithGivenFields` in DrupalContext.php.

I discovered that the  `role(s)` part of the annotation is the problem, specifically the parentheses, which I gather are messing up with the regex matching or something. Changing the annotation to:

`@Given /I am logged in as a user with the :role role and I have the following fields:`

and adjusting the corresponding step to remove the `(s)` allows it to work perfectly.

sensespidey:
… a user with roles.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/DrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/DrupalContext.php
@@ -35,7 +35,7 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext {
   /**
    * Creates and authenticates a user with the given role(s).
    *
-   * @Given I am logged in as a user with the :role role(s)
+   * @Given I am logged in as a user with the :role role
    * @Given I am logged in as a/an :role
    */
   public function assertAuthenticatedByRole($role) {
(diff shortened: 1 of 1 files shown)


pfrenssen:
This change will break every project that uses the plural form in their scenarios. 

I think if the matching of this step causes problems then it is something that needs to be fixed upstream in the Behat project. This pattern is allowed, it follows this example from the [Defining steps](https://docs.behat.org/en/latest/quick_start.html#defining-steps) documentation:

```
@Given there is a(n) :arg1, which costs £:arg2
```

sensespidey:
You're absolutely right, please ignore this! In hindsight, I was obviously confused about what was actually broken on the Behat feature I was working on, my apologies for the noise :)

PR #605, opened 2022-03-30
Add Behat\Gherkin\Node\TableNode namespace for instanceof TableNode

xurizaemon:
At line 67, we currently check `instanceof Tablenode`, but since the class isn't introduced with `use` we check for `instanceof Drupal\DrupalExtension\Context` which doesn't exist.

## Steps to reproduce

This should create a user with a random name, whose email is the same random value @example.org.

```feature
Given users:
  | name | mail |
  | <?random> | <?random>@example.org |
```

## Proposed solution

1. Add the `use` statement so we detect and populate random variables from TableNode data
2. Transform TableNode input data so that the variables are used


RoSk0:
I can confirm that proposed change is fixing random token handling in the tables.

CI fails with code style checks on unrelated to this PR files:
```
FILE: ...s/behat_test/src/Plugin/Field/FieldWidget/AddressFieldWidget.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
41 | ERROR | [x] Use null coalesce operator instead of ternary
    |       |     operator.
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------

FILE: ...s/behat_test/src/Plugin/Field/FieldWidget/AddressFieldWidget.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
41 | ERROR | [x] Use null coalesce operator instead of ternary
    |       |     operator.
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

@xurizaemon , can you please rebase your branch and push again. Maybe it will pass this time?

RoSk0:
It looks like a tested it poorly at first . 

While adding namespace is definitely required, it doesn't fix the problem. Random token does get detected with namespace in place in `beforeScenarioSetVariables()`, but `transformVariables()` does not get called for the steps with tables. So, the page created in https://github.com/jhedstrom/drupalextension/pull/633 test actually has `<?random_page>` title.

xurizaemon:
Updated, but I think your second comment is now saying this fix doesn't work?

Pulled in test from your repo also, thanks.

PR #602, opened 2022-01-06
WIP PHP 8.1 compatibility.

FlorentTorregrosa:
Hello,

I tried to update my skeleton on PHP 8.1 https://gitlab.com/florenttorregrosa-drupal/docker-drupal-project/-/merge_requests/107 and Behat is not working.

I tried to debug so here is a PR but encountered other problems.

First:
```
Run behat tests.

Deprecated: Method ReflectionParameter::getClass() is deprecated in /project/vendor/behat/mink-extension/src/Behat/MinkExtension/ServiceContainer/Driver/GoutteFactory.php on line 119

Deprecated: Method ReflectionParameter::getClass() is deprecated in /project/vendor/behat/mink-extension/src/Behat/MinkExtension/ServiceContainer/Driver/GoutteFactory.php on line 119

Deprecated: Return type of GuzzleHttp\Cookie\CookieJar::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /project/vendor/guzzlehttp/guzzle/src/Cookie/CookieJar.php on line 220

Deprecated: Return type of GuzzleHttp\Cookie\CookieJar::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /project/vendor/guzzlehttp/guzzle/src/Cookie/CookieJar.php on line 225
Feature: Global Elements

  ┌─ @BeforeScenario # Drupal\DrupalExtension\Context\MailContext::disableMail()
  │
  ╳  8192: strtolower(): Passing null to parameter #1 ($string) of type string is deprecated in /project/vendor/drupal/drupal-extension/src/Drupal/DrupalDriverManager.php line 71
  │
  @api
  Scenario: Homepage Contact Link                                                    # features/global.feature:4
    Given I am on the homepage                                                       # Drupal\DrupalExtension\Context\MinkContext::iAmOnHomepage()
    Then I should see the link "Home" in the "primary_menu" region                   # Drupal\DrupalExtension\Context\MinkContext::assertLinkRegion()
    Then I should see the link "Log in" in the "secondary_menu" region               # Drupal\DrupalExtension\Context\MinkContext::assertLinkRegion()
    Then I should see the link "Contact" in the "footer_fifth" region                # Drupal\DrupalExtension\Context\MinkContext::assertLinkRegion()
[...]

PR #601, opened 2021-11-14
#600: Avoid whitespace in tags 

andriokha:
Since Gherkin 4.9.0 tags with whitespace are deprecated: use a comma as an explicit OR operator.

See
- https://github.com/Behat/Gherkin/issues/213
- https://github.com/Behat/Gherkin/pull/215
- https://behat.org/en/latest/user_guide/context/hooks.html#tagged-hooks

Closes #600

Thanks!

PR #606, opened 2022-04-24
Php81

Berdir:
This is based on https://github.com/jhedstrom/drupalextension/pull/602, phpspec is now available and passes with this, but I suspect this is only the start.

Specifically, I expect we need to replace goutte with browserkit driver along the lines of https://www.drupal.org/project/drupal/issues/3176655. And the guzzle deprecation messages require guzzle 7, which in turn requires https://www.drupal.org/project/drupal/issues/3225966.


Berdir:
Ok, maybe we don't have to replace goutte just yet. That's also blocked on https://github.com/FriendsOfBehat/MinkExtension/pull/14.

instead, switching to the friends-of-behat minkextension fork, which seems to be php8.1 compatible. Still leaves the guzzle problem related to the core issue, but switching to v7 is not just blocked on that but also getting rid of fabpot/goutte, so again the issue above.

also some php8.1 fixes and a wrong config ID.

However, I am _very_ confused about the url generator error. I see several weird things around that, including the fact that optional service methods apparently do not seem properly supported by core (it should not call it with NULL but not call at all). Making it non-required causes deprecated warnings that need changes in DrupalDriver. But I don't yet see how that's 8.1 related.

I also included #601 to be able to run tests.

Berdir:
So I was able to fix the url generator problem  in drupal-driver.... by removing a few line of code, specifically just one would have been sufficient but I thought I'd remove them all as they didn't seem to cause a problem when not there. I guess that's all leftover code from a very, very long time ago or just copied from D7 without adjusting.

Merge request for that here: https://github.com/jhedstrom/DrupalDriver/pull/245

did a hack to include that commit here, somehow that only worked on D9, but there it passed: https://app.travis-ci.com/github/jhedstrom/drupalextension/jobs/568139241

What apparently happened is that this caused a php deprecation notice, which behat error handling converted to an exception and then that resulted in skipping that service and injecting NULL instead, which is a bug too but only a follow-up bug.

The guzzle php deprecation still happens, but it doesn't prevent tests from running, that will need to be patched for now or so, we can do a new MR to user browserkit directly once https://github.com/FriendsOfBehat/MinkExtension/pull/14 lands, then you can possibly update to guzzle 7 with some composer alias trickery or when https://www.drupal.org/project/drupal/issues/3225966 gets in.

composer.json
```diff
@@ -23,7 +23,7 @@
   "require": {
     "behat/behat": "~3.2",
     "behat/mink": "~1.5",
-    "behat/mink-extension": "~2.0",
+    "friends-of-behat/mink-extension": "^2",
     "behat/mink-goutte-driver": "~1.0",
```
claudiu-cristea, line 27: I need also `behat/mink-goutte-driver:~2.0`. Could we update to `~1.0 || ~2.0`

Berdir: I *think* I tried that, but don't remember why I didn't use that, possibly some conflict? 

this was merged, so will require a new MR anyway. That said, I think the path forward is to use browserkit directly once https://github.com/FriendsOfBehat/MinkExtension/pull/14 lands. Per https://github.com/minkphp/MinkGoutteDriver/releases/tag/v2.0.0, that's just a thin wrapper and you're supposed to switch over.

claudiu-cristea: @Berdir  I've created #610 (didn't test locally)

claudiu-cristea: @Berdir this is blocking me adopting PHP 8.1

> (...) that's just a thin wrapper and you're supposed to switch over.

To switch where? As Drupal Extension have this dependency, I suppose I need to switch here. But I can't find where

Berdir: the browserkit driver, as described in the 2.0.0 release notes: https://github.com/minkphp/MinkGoutteDriver/releases/tag/v2.0.0. And https://github.com/FriendsOfBehat/MinkExtension/pull/14 was just merged and released two days ago, so that might be possible now.

I guess that will require changes in this project, but it seems to be the mid/longterm path forward.

claudiu-cristea: @Berdir, what you're saying is that we should replace `behat/mink-goutte-driver` with `behat/mink-browserkit-driver` because the former is only a wrapper around the latter. Well, I did this in https://github.com/jhedstrom/drupalextension/pull/610 but it seems goutte driver is still requested by `friends-of-behat/mink-extension` as you can see in the test failure. I don't see anything we can do here, in drupal-extension

Berdir: > but it seems goutte driver is still requested by friends-of-behat/mink-extension as you can see in the test failure.

friends-of-behat/mink-extension only has a require-dev dependency on goutte-driver.

Not sure what's the earlier and the current MR, but the test fails I see currently in the MR are because of conflicts with older PHP versions (symfony 6 requires PHP 8.1) and drupal 9 vs symfony versions conflicts. Which makes sense, my guess is that this approach won't work until a new major version that can require Drupal 10 or something.

The current state works for me on php 8.1 beside those initial guzzle deprecation messages. I'm not against requiring or maybe additional supporting v2 so that you can pick a matching version depending on drupal and php version, just that it's not a long-term solution.

PR #611, opened 2022-06-27
Fix drupal-extension requirement

zagrad:
With version 3.x the `bin/behat --init` command fails.


AlexSkrypnyk:
Thank you for your contribution @zagrad 
Similar update was made https://github.com/jhedstrom/drupalextension/blob/main/doc/_static/snippets/composer.json#L3

PR #613, opened 2022-07-13
Drupal core default theme changed to Olivero

claudiu-cristea:
Some strings on the page were changed in Drupal 9.4


jhedstrom:
Thanks!

PR #610, opened 2022-06-23
behat/mink-goutte-driver update

claudiu-cristea:
Allow also ~2.0 of behat/mink-goutte-driver

PR #615, opened 2022-08-04
Fixes #614 for PHP 8.1, Symfony 6, Drush 11, etc.

### LINKED ISSUE #614: Drupal 10 / Drush 11 / Symfony 6 Support
I'm trying to do some testing with PHP 8.1 and I'm getting blocked by some of the requirements with regards to symfony on this package. We need to make sure it allows symfony ^6 packages. I'll open a PR for review!

mikemadison13:
Updates the following:

- drupal related dependencies to allow 10.x
- drush to allow drush 11
- behat/mink-goutte-driver to allow 2.x
- symfony packages to allow ^6

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,9 +28,9 @@
     "behat/mink-goutte-driver": "~1|^2",
     "behat/mink-selenium2-driver": "~1.1",
     "drupal/drupal-driver": "^2.1.0",
-    "symfony/browser-kit": "^3.4|~4.4|^6",
-    "symfony/dependency-injection": "~3.0|~4.4|^6",
-    "symfony/translation": "^3.4|~4.4|^6"
+    "symfony/browser-kit": "~4.4|^6",
+    "symfony/dependency-injection": "~4.4|^6",
+    "symfony/translation": "~4.4|^6",
   },
   "require-dev": {
     "composer/installers": "^1.2|^2",
--- a/composer.json
+++ b/composer.json
@@ -24,13 +24,13 @@
   "require": {
     "behat/behat": "~3.2",
     "behat/mink": "~1.5",
-    "friends-of-behat/mink-extension": "^2",
+    "friends-of-behat/mink-extension": "^2.7.2",
     "behat/mink-goutte-driver": "~1|^2",
     "behat/mink-selenium2-driver": "~1.1",
     "drupal/drupal-driver": "^2.1.0",
     "symfony/browser-kit": "~4.4|^6",
     "symfony/dependency-injection": "~4.4|^6",
-    "symfony/translation": "~4.4|^6",
+    "symfony/translation": "~4.4|^6"
   },
   "require-dev": {
     "composer/installers": "^1.2|^2",
--- a/composer.json
+++ b/composer.json
@@ -20,23 +20,24 @@
     }
   ],
   "prefer-stable": true,
+  "minimum-stability": "dev",
   "require": {
     "behat/behat": "~3.2",
     "behat/mink": "~1.5",
     "friends-of-behat/mink-extension": "^2",
-    "behat/mink-goutte-driver": "~1.0",
+    "behat/mink-goutte-driver": "~1|^2",
     "behat/mink-selenium2-driver": "~1.1",
     "drupal/drupal-driver": "^2.1.0",
-    "symfony/browser-kit": "^3.4|~4.4",
-    "symfony/dependency-injection": "~3.0|~4.4",
-    "symfony/translation": "^3.4|~4.4"
+    "symfony/browser-kit": "^3.4|~4.4|^6",
+    "symfony/dependency-injection": "~3.0|~4.4|^6",
+    "symfony/translation": "^3.4|~4.4|^6"
   },
   "require-dev": {
-    "composer/installers": "^1.2",
+    "composer/installers": "^1.2|^2",
     "drupal/coder": "^8.3",
-    "drupal/core-composer-scaffold": "^9.1",
-    "drupal/core-recommended": "^9.1",
-    "drush/drush": "^10.5",
+    "drupal/core-composer-scaffold": "^9.1|^10.0.0-alpha1",
+    "drupal/core-recommended": "^9.1|^10.0.0-alpha1",
+    "drush/drush": "^10.5|^11 ",
     "php-parallel-lint/php-parallel-lint": "^1.3",
     "phpspec/phpspec": "^4.0 || ^6.0 || ^7.0"
   },
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 7.3
   - 7.4
   - 8.0
   - 8.1
@@ -11,7 +10,6 @@ env:
     - PATH=$PATH:/home/travis/.config/composer/vendor/bin
     - TRAVIS_NODE_VERSION="4.0.0"
   matrix:
-    - DRUPAL_VERSION=7
     - DRUPAL_VERSION=9
 
 # Enable Travis containers.
@@ -28,17 +26,13 @@ install:
   # @todo Re-enable behat drush endpoint testing.
   # @see https://github.com/jhedstrom/drupalextension/issues/458
   - test ${DRUPAL_VERSION} -lt 8 || composer require --no-interaction --dev --no-update drupal/core-recommended:^${DRUPAL_VERSION} drupal/core-composer-scaffold:^${DRUPAL_VERSION} drush/drush:^10
-  # Remove core/drush deps for Drupal 7.
-  - test ${DRUPAL_VERSION} -ge 8 || composer remove --dev drush/drush drupal/core-recommended drupal/core-composer-scaffold
   - composer install
   # For Drupal 8, change composer.json to require symfony 3.x as conflicting
   # versions between the separate drupal and drupalextension vendor folders
   # can create conflicts.
   # Install drush globally.
   - (test ${DRUPAL_VERSION} -lt 8 && composer global require drush/drush:~8.0 drupal/drupal-driver) || composer global require drush/drush:~10.0
-  # Install the Behat Drush Endpoint for Drupal 7 tests.
-  - test ${DRUPAL_VERSION} -gt 7 || (git clone https://github.com/drush-ops/behat-drush-endpoint.git drush/behat-drush-endpoint && (cd drush/behat-drush-endpoint && composer install --no-dev && cd -))
-  # Pin node version.
+   # Pin node version.
   # @see http://austinpray.com/ops/2015/09/20/change-travis-node-version.html
   - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION
   - npm install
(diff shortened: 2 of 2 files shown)


Berdir:
Overlaps with https://github.com/jhedstrom/drupalextension/pull/610

mikemadison13:
@Berdir yeah, i know it overlaps with a couple of other efforts. i'm hoping we can get at least one of them merged, though. my behat testing is currently blocking me from being able to execute CI on my PHP 8.1 project update.

mikemadison13:
the failure I'm hitting seems to be related to https://github.com/symfony/panther/issues/291, digging more.

mikemadison13:
@jhedstrom digging into the issues in the build failure, it looks like https://github.com/FriendsOfBehat/MinkExtension did some refactoring in version 2 which is causing the set client error in the build:

```
In GoutteFactory.php line 117:
                                                    
Method Goutte\Client::setClient() does not exist  
```

I've started poking at this a little bit, it looks like they are checking for Goutte1 in their code (see https://github.com/FriendsOfBehat/MinkExtension/blob/master/src/Behat/MinkExtension/ServiceContainer/Driver/GoutteFactory.php#L115) which could be the root cause.

I'm going to go open an issue over there to see if we can make forward progress.

mikemadison13:
actually, it looks like there is a PR that fixes this https://github.com/FriendsOfBehat/MinkExtension/pull/16 and it works, but not for the Drupal 7 stuff. I think we may need to split the drupalextension and have one version that is for newer versions of symfony. any thoughts?

jhedstrom:
@mikemadison13 yeah, I think we could start a 5.x and drop support for Drupal 7 in that branch?

mikemadison13:
it's unclear to me if there's anything in #610 that we need to roll into this. i don't think so based on what i see there, but that PR does change the behat.yml.dist file from goutte to browserkit_http. i'm not sure if that's still needed? @claudiu-cristea might have some input?

mikemadison13:
i think we can also close #602 as well given this?

composer.json
```diff
@@ -59,6 +61,11 @@
     }
   },
   "extra": {
+    "patches": {
+        "friends-of-behat/mink-extension": {
+            "setClient error": "https://github.com/FriendsOfBehat/MinkExtension/files/9480850/16.txt"
```
jhedstrom, line 66: Let's add the PR number `#16` here for easy reference to the issue in case we forget where this came from :)

composer.json
```diff
@@ -59,6 +61,11 @@
...
+        }
+    },
     "branch-alias": {
       "dev-master": "4.3.x-dev"
```
jhedstrom, line 70: you'll want to merge `master` in since this line has changed for `5.0.x` now.

claudiu-cristea:
@mikemadison13 I don't have time for #610 and, honestly, I have no idea how to fix it. Feel free to include here anything that could be useful

claudiu-cristea:
@mikemadison13 This PR is fixing the dependency issues I have when updating to PHP 8.1. Thank you. Could you, please fix the 2 minor remarks and let's get this in.

composer.json
claudiu-cristea: @jhedstrom In my project, this patch failed to apply, I had to add this ignore:

```
"patches-ignore": {
  "drupal/drupal-extension": {
    "friends-of-behat/mink-extension": {
      "setClient error": "https://github.com/FriendsOfBehat/MinkExtension/files/9480850/16.txt"
    }
  }
},
```

`friends-of-behat/mink-extension` 2.3.1 is installed

claudiu-cristea: I created https://github.com/mikemadison13/drupalextension/pull/1 (against fork)

composer.json
claudiu-cristea: I created https://github.com/mikemadison13/drupalextension/pull/1 (against fork)

composer.json
claudiu-cristea: In my project the patch doesn't apply, friends-of-behat/mink-extension is locked to 2.3.1

mikemadison13: i wonder if we need to set a minimum version to 2.7.1 of the mink-extension? I'm not sure how we make this PR work without that patch (because it fatal errs out). So we either need to pin way earlier (2.3.1 and remove the patch) or pin later (2.7.1) with. my preference would be the newer version. thoughts @jhedstrom ?

composer.json
```diff
@@ -59,6 +61,11 @@
     }
   },
   "extra": {
+    "patches": {
+        "friends-of-behat/mink-extension": {
+            "setClient error from PR16": "https://github.com/FriendsOfBehat/MinkExtension/files/9480850/16.txt"
```
Berdir, line 66: even with the version constraint, a patch is IMHO too unstable, sooner or later that won't apply anymore or will be merged. We could maybe combine it with a specific requirement on that specific version, but waiting for that to be merged and committed is probably the safer option.

mikemadison13: @Berdir there have been a number of requests to merge the PR and it hasn't happened, so I don't fundamentally disagree with the concern, but I'm not sure if the maintainer can merge things at the moment given their location. we did try and make this "more stable" by pointing at a patch file vs. the actual PR. i guess we could explicitly pin to 2.7.1 to prevent accidental breakage / updates?

AlexSkrypnyk:
Can we please get this merged. The CI passes, the patch is actually a "static" file - all looks like it is working.

If that patch won't apply - we can always roll a new patch version of this package.

This is currently blocks using Behat with Drupal 10 big time.

claudiu-cristea:
@jhedstrom @Berdir @mikemadison13 wouldn't be better if we try to move away from the Goutte driver? I see here how Moodle community did it https://github.com/stronk7/moodle/compare/master...bump_goutte_driver_20

Berdir:
Sure, that's what I always expected will need to happen eventually, see first comment in https://github.com/jhedstrom/drupalextension/pull/606 :) Back then it was blocked on https://github.com/FriendsOfBehat/MinkExtension/pull/14, but now that should be possible?

claudiu-cristea:
> Sure, that's what I always expected will need to happen eventually, see first comment in #606 :) Back then it was blocked on [FriendsOfBehat/MinkExtension#14](https://github.com/FriendsOfBehat/MinkExtension/pull/14), but now that should be possible?

@Berdir could you, please, take a loot at #620. It's not ready yet but I have a green there

claudiu-cristea:
@AlexSkrypnyk could you take a look and try #620? That drops Goutte Driver.

@mikemadison13  should we close this in favor of #620?

claudiu-cristea:
> my strong preference would be to get this in ASAP so we can unblock d10

@mikemadison13 we're already there. See the tests on GitHub actions here https://github.com/claudiu-cristea/drupalextension/actions/runs/3266111531. I'm now investigating the failure with Drupal 10, which is happening in drupal/drupal-driver

mikemadison13:
i believe we would still need to merge in all of the symfony 6 stuff from this (or another PR) to get us to d10, right?

claudiu-cristea:
> i believe we would still need to merge in all of the symfony 6 stuff from this (or another PR) to get us to d10, right?

I've cherry picked some commits from here in #620. Actually, it supports Sf 6, Drush 11. I'm still working on D10 stuff. Very close...

AlexSkrypnyk:
The upstream blocker was merged.

If I may suggest - can we update this PR, merge it, release it and then look at 620?

claudiu-cristea:
@AlexSkrypnyk they are doing the same, with the exception that https://github.com/FriendsOfBehat/MinkExtension/pull/16 wraps the BrowserKit Driver from Goutte Driver while #620 uses directly the BrowserKit Driver and drops the dependency to Goutte Driver

AlexSkrypnyk:
@jhedstrom 
Can we please merge and release this one and tackle #620. There are already several complex dependency resolution issues happening here. Bringing in a new dependency in this PR (which is 100% working and is ready to go) can potentially throw us back into discussions about approaches. I would suggest to have those discussions separately (in #620) and merge this one as is.

claudiu-cristea:
@AlexSkrypnyk I don't see why this needs to be merged as #620 is solving exactly the same issue but as a definitive fix. Did you give #620 a try? If we merge this just to drop it in #620 somebody has to rework/reroll #620. Useless work. Why not, better, focusing on that to se if fulfills the scope?

AlexSkrypnyk:
@claudiu-cristea 
That PR adds a lot of changes that need to be discussed. I simply want to unblock downstream packages ASAP and then take time to review #620.

There is only 1 files change in this PR (`composer.json`) that will need a re-roll in #620.

Berdir:
goutte v2 is just a wrapper around browserkit, I agree that it makes more sense to pursue that directly.

claudiu-cristea:
@AlexSkrypnyk 90% of changes are only file removals, renaming, docs changes and Behat tags removals because it drops the Drupal 7/8 support

mikemadison13:
just to chime in to support the point @AlexSkrypnyk is making, it would be fantastic to get D10 unblocked asap. if that's with this pr great. if that's with 620 great, i don't have a lot of skin in the game. i just want to get behat unblocked. it feels safer to  use the current approach first and then explore a new approach separately, but 🤷

AlexSkrypnyk:
@Berdir @claudiu-cristea 
It makes sense to replace goutte as it is a wrapper around browserkit and should be decommissioned. But it is about when to do it and how long it will take to review and accept changes in #620 (which I think are great).

#620 has a lot of changes: there are new local env configs, new CI configs, Drupal 8 deprecations, test artefacts and other changes that will take time to discuss. I guess it is up to project maintainer to decide if changes in that PR can be accepted as-is or more discussions are required.

We have this PR that is working and does not change any dependencies (so that downstream projects do not have to update theirs). It is ready to be merged right now. It has minimal changes and, therefore, less risk to introduce new issues.

I do not see a problem with merging this PR as-is and releasing a new version (could even be a patch version) to unblock downstream and then concentrate on #620 and spend as much time as required there.

jhedstrom:
This could potentially go against the latest 4.x.x branch, and keep #620 against `master` (which will be 5.0.x)?

mikemadison13:
@jhedstrom i think we discussed cutting the 5.x branch for Drupal 10 given that we had to remove some of the earlier supported versions of things, right? just making sure i don't misremember. this branch will not work well with older things.

jhedstrom:
Yeah, 5.x will drop D7 support and add D10 support, so scratch my thought above. Since 5.x is not released, and #620 is the eventual direction we'd want to go anyway, I'm inclined to go that route...

mikemadison13:
cool, if that's the route to go then as i said above i don't really care. i just would like the most expedient route to d10. so if we aren't merging this and cutting it (despite it being ready to go right now) when can we have the other? happy to help test once we have the dependencies merged.

claudiu-cristea:
I guess this can be closed. The fix from #620 has been already merged in `master`. I would love to get some feedback from projects using the current `master`.

jhedstrom:
Thanks for all the work here! I'm closing this out since #620 was merged.

AlexSkrypnyk:
@jhedstrom
This PR was authored in October 2022. It was then closed in Dec 2022. It is now March 2023 and we still do not have a stable release with this update for `4.x` or `5.x` after 5 months :(

The reason for submitting this pull request was to prevent this specific situation from occurring.

Could you please advise what is the planned course of actions here? Is 5.x going to have a release soon? Can this PR be ported to `4.x` and released?

jhedstrom:
@AlexSkrypnyk thanks for the reminder. Since no major issues have been reported with the alpha release I've tagged RC1 for 5.0.0.

PR #620, opened 2022-10-15
Drop Goutte Driver. Support D10, Drush 11, Symfony 6

claudiu-cristea:
Tests at https://github.com/claudiu-cristea/drupalextension/actions/


claudiu-cristea:
@jhedstrom any idea why the tests are not running with the last commit?

Berdir:
I see that 5.x is open as development version, makes sense to introduce the hard break here. Won't have time for a closer look in the next week or so, but looks great at first glance.

claudiu-cristea:
@Berdir unfortunately applied to my project I hit this (which I experienced also with the #615 https://github.com/symfony/symfony/pull/47505. I really think that it’s a symfony/mime bug. And it happens a lot with Drupal forms, just think at forms with files, i.e. `file[0]` fields

claudiu-cristea:
@jhedstrom @Berdir @mikemadison13 we have green with GitHub actions:

* You can see the tests on my repo https://github.com/claudiu-cristea/drupalextension/actions/runs/3266548301/jobs/5370436816. Somehow, the main repo should be configured to run here instead of Travis CI
* Backwards compatibility for Drupal 7/8 is dropped. This should go in a `5.0.x` branch.
* Didn't delete yet the `.travis.yml` file. Needs @jhedstrom decision on whether we move to GitHub actions
* On my project I'm experiencing this issue with browser kit driver https://github.com/symfony/symfony/issues/47504 (PR https://github.com/symfony/symfony/pull/47505). Will try to add a test here to see if I can reproduce.

pfrenssen:
Replacing Travis with Github actions makes sense imo. And @jhedstrom also seems to be on board with this, ref https://github.com/jhedstrom/DrupalDriver/issues/254

mikemadison13:
do we have any updates here? if this is going to continue on, it would be awesome if we could merge #615 so we can unblock d10 and then do the feature development here. thanks so much!

claudiu-cristea:
@mikemadison13 I don't get why everybody wants to see #615 merged but avoid to give this a chance by running the tests. I've been tested with both on and both are having exactly the same effect, just that this one is a definitive fix.

mikemadison13:
totally a fair question. i always try to do feature development separate from maintenance, that's my big thing. a 100+ files changing in a PR feels like iterating on a feature / expanding capabilities, which is different activity from just updating dependencies to allow an install. i don't fundamentally care which PR gets merged, i'm just saying i'd like one of them to get merged soon (and if 615 is easier to test, then we should start there).

claudiu-cristea:
As I already told, the big amount of file changes is a result of dropping Drupal 7/8 support and fixing the documentation. Actually, #615 is also dropping the support but _forgets_ to remove the tests and fix the documentation. I also had to switch from Travis CI to GitHub Actions and the tests stop running because of some quota limit.

claudiu-cristea:
@mikemadison13 let's discuss here https://drupal.slack.com/archives/C1BMUQ9U6/p1665865781863389

claudiu-cristea:
Opened #622 as follow-up

composer.json
```diff
@@ -71,13 +69,21 @@
       "locations": {
         "web-root": "drupal/"
       }
+    },
+    "enable-patching": true,
+    "composer-exit-on-patch-failure": true,
+    "patches": {
+        "drupal/drupal-driver": {
+            "https://github.com/jhedstrom/DrupalDriver/pull/253": "https://patch-diff.githubusercontent.com/raw/jhedstrom/DrupalDriver/pull/253.diff"
```
jhedstrom, line 77: Instead of patching, I'll merge https://github.com/jhedstrom/DrupalDriver/pull/253 and then we can update composer.json to pull dev until we release 5.0.

jhedstrom:
This is in! Thanks all for sticking with this, and pushing it forward!

PR #622, opened 2022-11-07
Use BrowserKit with Guzzle client

claudiu-cristea:
This PR is built on top of #620.

## Problem

* #620 allows to get rid of Goutte driver and use directly BrowserKit driver
* However, with BrowserKit, which uses the HttpBrowser, the forms that are posting files are failing. The reason is this https://github.com/symfony/symfony/issues/47504 (at least I couldn't make it work)
* This issue could not be reproduced by [Drupal Test Traits](https://gitlab.com/weitzman/drupal-test-traits) because they are using `BrowserKit` with Guzzle as client, via `Drupal\Tests\DrupalTestBrowser`, see https://gitlab.com/weitzman/drupal-test-traits/-/blob/2.x/src/BrowserKitTrait.php#L21

### Proving the bug

I've created a _"test only"_ PR #623. See the failures here https://github.com/claudiu-cristea/drupalextension/actions/runs/3411428842

## Proposal

Use `Drupal\Tests\DrupalTestBrowser` as `BrowserKitDriver` client. This also has the advantage that is aligned with Drupal core PHPUnit tests.

Here's a diff to #620 https://github.com/claudiu-cristea/drupalextension/pull/1


claudiu-cristea:
Tests:

* Without this PR (fail) https://github.com/claudiu-cristea/drupalextension/actions/runs/3411442593/jobs/5675645228
* With this fix (pass) https://github.com/claudiu-cristea/drupalextension/actions/runs/3411442593

PR #627, opened 2022-11-16
Test image/file uploads

claudiu-cristea:


claudiu-cristea:
@jhedstrom I'm out of ideas. I cannot understand why tests are not running in the main repo but only on the forked repo (see https://github.com/claudiu-cristea/drupalextension/actions/runs/3478024729)

jhedstrom:
This was a test-only PR. Closing.

PR #628, opened 2022-11-16
Use BrowserKit with Guzzle client

claudiu-cristea:
## Problem

* We've switched from Goutte to BrowserKit driver in #620 
* However, with BrowserKit, which uses the HttpBrowser, the forms that are posting files are failing. The reason is this https://github.com/symfony/symfony/issues/47504
* This issue could not be reproduced by [Drupal Test Traits](https://gitlab.com/weitzman/drupal-test-traits) because they are using `BrowserKit` with Guzzle as client, via `Drupal\Tests\DrupalTestBrowser`, see https://gitlab.com/weitzman/drupal-test-traits/-/blob/2.x/src/BrowserKitTrait.php#L21

### Proving the bug

I've created a _"test only"_ PR #627

## Proposal

Use `Drupal\Tests\DrupalTestBrowser` as `BrowserKitDriver` client. This also has the advantage that is aligned with Drupal core PHPUnit tests.


claudiu-cristea:
Green tests are here https://github.com/claudiu-cristea/drupalextension/actions/runs/3478051951

jhedstrom:
This looks great, thanks!

PR #655, opened 2023-12-06
Make Guzzle request options configurable

claudiu-cristea:
## Problem

Behat tests are failing for logged in users when the site is accessed via HTTPS and SSL peer verification is failing. But with `BrowserTestBase` we're able to run such tests. Testing should not require a valid SSL certificate

## Proposal

Make Guzzle request options configurable

Example:

```yaml
default:
  suites:
    default:
      contexts:
        - FeatureContext
        - Drupal\DrupalExtension\Context\DrupalContext
        - Drupal\DrupalExtension\Context\MinkContext
  extensions:
    Drupal\MinkExtension:
      base_url: https://localhost
      browserkit_http:
        guzzle_request_options:
          verify: false
```

In this example, `browserkit_http.guzzle_request_options` is configured to instantiate the Guzzle underlying HTTP client with the `verify` request option set to `false`, same as `BrowserTestBase:: initMink()` is doing

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php_version: ["8.1", "8.2"]
-        drupal_version: ["10.0", "10.1"]
+        php_version: ["7.4", "8.0", "8.1", "8.2"]
+        drupal_version: ["9", "10.0", "10.1"]
+        exclude:
+          - php_version: "7.4"
+            drupal_version: "10.0"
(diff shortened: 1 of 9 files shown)


Berdir:
this should probably say *with self-signed certificate in the title, because that's the only issue here afaik, would need a setting to allow those I guess.

Also has  nothing to with this project, if not already supported it's something that the driver needs to support? I know some chrome drivers have this

claudiu-cristea:
Yes, maybe we should do the Guzzle request options configurable

claudiu-cristea:
Ready for review

PR #661, opened 2024-03-06
Allow creation of entities of any type with base field value parsing

### LINKED ISSUE #194: Given (a/an) :entity entity/entities/item/items

chrisolof:
Resolves #194.

This was built and tested in a D10 environment.  Pairs really well with jhedstrom/DrupalDriver/pull/271.

Opens up stuff like this:

```gherkin
Given "commerce_product_variation" entities:
  | title | type     | field_manufacturer | sku      | field_category | price:number | price:currency_code | status |
  | TNT   | product  | ACME               | TNT-01   | Excavation     | 200          | USD                 | 1      |
  | Anvil | product  | ACME               | ANVIL-01 | Metal          | 450          | USD                 | 1      |
  And "commerce_product" entities:
    | title | type     | variations | stores         | status |
    | TNT   | product  | TNT        | ACME Web Store | 1      |
    | Anvil | product  | Anvil      | ACME Web Store | 1      |
```

Deletes the entities out after scenario in reverse order so as to respect dependent entity setups.

Otherwise pretty much follows the node/term/user creation and cleanup patterns.  But now we can set up entities of _any_ type.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/DrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/DrupalContext.php
@@ -339,6 +339,23 @@ class DrupalContext extends RawDrupalContext implements TranslatableContext
         $this->assertSession()->statusCodeEquals('200');
     }
 
+  /**
+   * Creates entities of a given type provided in a table.
+   *
+   * Example:
+   * | title    | author     | status | created           |
+   * | My title | Joe Editor | 1      | 2014-10-17 8:00am |
+   * | ...      | ...        | ...    | ...               |
+   *
+   * @Given :type entities:
+   */
+    public function createEntities($type, TableNode $entitiesTable)
+    {
+        foreach ($entitiesTable->getHash() as $entityHash) {
+            $entity = (object) $entityHash;
+            $this->entityCreate($type, $entity);
+        }
+    }
 
   /**
    * Creates a term on an existing vocabulary.
--- a/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
@@ -73,6 +73,17 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface
     protected $terms = [];
 
   /**
+   * Keep track of created entities so they can be cleaned up.
+   *
+   * Each entry is expected to have the following keys:
+   * - type: The entity type.
+   * - entity: The saved entity object.
+   *
+   * @var array
+   */
+    protected $entities = [];
+
+  /**
    * Keep track of any roles that are created so they can easily be removed.
    *
    * @var array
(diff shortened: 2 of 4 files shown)


DamienGR:
Thank you ! I spent few hours trying to make the price field works and could not understand why it worked for other compound fields (like address field). The base fields were indeed ignored from the parsing method

AlexSkrypnyk:
Hey @chrisolof - thanks for this contribution! It's been a while since the last update here, so I've picked up the work and iterated on it.

The driver API moved to a typed `EntityStubInterface` since this branch was opened, so I've reimplemented the generic step against the current contracts in #853. The good news is that the cleanup machinery you outlined here has since landed in `RawDrupalContext::cleanEntities()` and `deleteStub()` (which already routes unknown entity types to `$driver->entityDelete($stub)`), so the new PR is just the public step, the generic `entityCreate()`, and the `Before/AfterEntityCreate` hook scopes/attributes for parity with the per-type hooks. Your work is attributed via a `Co-authored-by` trailer.

I'll close this PR once #853 lands. Appreciate the initial work and the discussion that made it concrete!

AlexSkrypnyk:
Landed via #853. Thanks again, @chrisolof!

chrisolof:
Thanks @AlexSkrypnyk!  I dove into the current codebase this morning to see how you are achieving things like compound base field parts parsing (handling that price field example above) and entity deletion in the reverse-order of creation (both of which I rely on quite a bit) and it looks great!  I also love the thorough test coverage over all of this functionality.  Beyond just proving all of this works as designed, that test coverage will continually provide users like myself up-to-date, working examples of how to use this system correctly.

I'm excited to test this out in my next round of dependencies updates.

Thanks for all of your work on this project (and the related drupal/drupal-driver project), and for incorporating all of these long-standing PRs!

PR #663, opened 2024-03-06
Add step definition for dragging one element onto another

### LINKED ISSUE #662: Add step definition for dragging one element onto another
We currently use a custom step like this for testing through draggable interfaces:

```gherkin
Given I drag element :dragged onto element :target
```

This seems like a step definition that would be useful to many projects (not just ours), so I'd like to contribute it back.  PR to follow.

chrisolof:
Resolves #662.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MinkContext.php
+++ b/src/Drupal/DrupalExtension/Context/MinkContext.php
@@ -3,6 +3,7 @@
 namespace Drupal\DrupalExtension\Context;
 
 use Behat\Behat\Context\TranslatableContext;
+use Behat\Mink\Exception\ElementNotFoundException;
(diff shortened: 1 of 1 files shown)


AlexSkrypnyk:
Hey @chrisolof - thanks for this contribution! It's been a while since the last update here, so I've picked up the work and iterated on it.

I've opened #800 which builds on your changes with some updates to bring it in line with the current codebase - PHP 8 attributes, type declarations, changed the annotation from `@Given` to `@When` (since it's an action), and added Behat tests with an HTML fixture for drag-and-drop. Your commit is preserved with original authorship.

I'll close this PR once the new one is merged. Appreciate the initial work!

AlexSkrypnyk:
Landed via #800. Thanks again, @chrisolof!

PR #664, opened 2024-03-11
Do not set the globally installed Drush as default binary path

mxr576:
Extra handling in connection with https://github.com/jhedstrom/DrupalDriver/issues/272 and the proposed change in https://github.com/jhedstrom/DrupalDriver/issues/273.

### DIFF
--- a/src/Drupal/DrupalExtension/ServiceContainer/DrupalExtension.php
+++ b/src/Drupal/DrupalExtension/ServiceContainer/DrupalExtension.php
@@ -264,7 +264,7 @@ class DrupalExtension implements ExtensionInterface
             $config['drush']['alias'] = isset($config['drush']['alias']) ? $config['drush']['alias'] : false;
             $container->setParameter('drupal.driver.drush.alias', $config['drush']['alias']);
 
-            $config['drush']['binary'] = isset($config['drush']['binary']) ? $config['drush']['binary'] : 'drush';
+            $config['drush']['binary'] = isset($config['drush']['binary']) ? $config['drush']['binary'] : '';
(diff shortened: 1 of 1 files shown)


src/Drupal/DrupalExtension/ServiceContainer/DrupalExtension.php
```diff
@@ -264,7 +264,7 @@ private function loadDrush(FileLoader $loader, ContainerBuilder $container, arra
             $config['drush']['alias'] = isset($config['drush']['alias']) ? $config['drush']['alias'] : false;
             $container->setParameter('drupal.driver.drush.alias', $config['drush']['alias']);
 
-            $config['drush']['binary'] = isset($config['drush']['binary']) ? $config['drush']['binary'] : 'drush';
+            $config['drush']['binary'] = isset($config['drush']['binary']) ? $config['drush']['binary'] : '';
```
AlexSkrypnyk: I think this should throw a `RuntimeException` if the binary path is not set. alternatively, it may default to `vendor/bin/drush`. But leaving this empty could trigger a cascade of issues that will be hard to debug.

AlexSkrypnyk (changes requested):

AlexSkrypnyk:
Hey @mxr576 — thanks for raising this. You were right that defaulting to a global `drush` is problematic.

I've opened #799 which takes a slightly different approach — defaulting to `vendor/bin/drush` instead of an empty string, so it works out of the box for Composer-managed projects without risking the cascade of issues an empty path would cause.

Closing this one in favour of that approach. Appreciate you flagging it!

Issue #557, opened 2019-10-22
iWaitForAjaxToFinish() fails


brooke-heaton:
I'm consistently getting null `$result` on `drupal-extension/src/Drupal/DrupalExtension/Context/MinkContext.php:156` I'm not clear on where the failure is but this essentially makes any `@javascript` Feature impossible. I'm on Drupal v `8.7.8` using Lando with the `selenium/standalone-chrome:3.141.59-oxygen` chromedriver on PHP 7.2

Full error output for reference: 

```
[Testing\Behat] Running behat  --format pretty /app/tests/behat --colors --no-interaction --stop-on-failure --strict --config /app/tests/behat/naswa.behat.yml --profile naswa --tags @conference -v
@naswa @javascript @api @release @fields @conference
Feature: Conference
  In order to verify that the conference has the correct fields and form display
  As an administrator
  I should be able to go to the conference create page and all fields

  Scenario: Create a conference page with all fields                      # features/naswa/release/ContentEditing/ConferenceFields.feature:7
    Given I am logged in as user "bheaton@naswa.org"                      # Drupal\FeatureContext::iAmLoggedInAsUser()
    Given I am on "/node/add/conference"                                  # Drupal\DrupalExtension\Context\MinkContext::visit()
    ┌─ @BeforeStep # Drupal\DrupalExtension\Context\MinkContext::beforeJavascriptStep()
    │
    ╳  RuntimeException: Unable to complete AJAX request. {"name":"step.before","feature":"Conference","step":"I click the \"Primary Info\" tab","suite":"default"} in /app/vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/MinkContext.php:173
    ╳  Stack trace:
    ╳  #0 /app/vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/Context/MinkContext.php(107): Drupal\DrupalExtension\Context\MinkContext->iWaitForAjaxToFinish(Object(Behat\Behat\Hook\Scope\BeforeStepScope))
    ╳  #1 [internal function]: Drupal\DrupalExtension\Context\MinkContext->beforeJavascriptStep(Object(Behat\Behat\Hook\Scope\BeforeStepScope))
    ╳  #2 /app/vendor/behat/behat/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php(109): call_user_func_array(Array, Array)
[...]

PR #666, opened 2024-04-17
Issue #557: iWaitForAjaxToFinish fails

AliagaDev:
jQuery.active internal value may have negative value in some cases causing inconsistent test results, as jquery supposedly increments the value when an ajax request starts and decrements it when it finishes, it should not influence the identification of active ajax requests


rsanzante (approved):
LGTM as well.

PR #671, opened 2024-07-04
Fix logout in drupal 10.3.x

ol0lll:
A new logout confirmation step was introduced in drupal 10.3.x, see https://www.drupal.org/project/drupal/issues/144538.

### DIFF
--- a/src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
+++ b/src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
@@ -103,6 +103,7 @@ class DrupalAuthenticationManager implements DrupalAuthenticationManagerInterfac
     public function logout()
     {
         $this->getSession()->visit($this->locatePath($this->getDrupalText('logout_url')));
+        $this->getSession()->getPage()->findButton($this->getDrupalText('log_out'))->click();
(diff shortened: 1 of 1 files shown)


src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
```diff
@@ -103,6 +103,7 @@ public function logIn(\stdClass $user)
     public function logout()
     {
         $this->getSession()->visit($this->locatePath($this->getDrupalText('logout_url')));
+        $this->getSession()->getPage()->findButton($this->getDrupalText('log_out'))->click();
```
eiriksm, line 106: Maybe we should just check if the element exist first? So that it's compatible with versions before 10.3 and after?

eiriksm (changes requested):

tame4tex:
See alternate solution in [#681]

src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
tame4tex: I have addressed this in an alternate solution [#681]

jhedstrom:
Thanks all! #681 is merged.

PR #673, opened 2024-07-09
Add ability to check for expected email attachments

### LINKED ISSUE #672: Add the ability to check for expected email attachments in MailContext
I would like to be able to test for the presence of an expected email attachment, in addition to correct `to` and `subject` fields.  For example:

```gherkin
  ...
  Then an email has been sent to "test@example.com" with the subject "Foo" and the attachment "bar.pdf"
  ...
```

Or, in the case of multiple attachments:

```gherkin
  ...
  Then an email has been sent to "test@example.com" with the subject "Foo" and the attachments "bar1.pdf,bar2.pdf"
  ...
```

I intend to follow this up with a pull request.

chrisolof:
Resolves #672.

Mail attachments seem to live at `$mail['params']['attachments']`.  Test coverage is not included due to Drupal\DrupalExtension\Context\MailContext::drupalSendsMail() relying on Drupal\Driver\DriverInterface::sendMail(), which appears to have no support for the direct inclusion of attachments, nor support for adding mail parameters like 'attachments'.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/MailContext.php
+++ b/src/Drupal/DrupalExtension/Context/MailContext.php
@@ -87,11 +87,14 @@ class MailContext extends RawMailContext
    * @Then (a )(an )(e)mail(s) has/have been sent to :to:
    * @Then (a )(an )(e)mail(s) has/have been sent with the subject :subject:
    * @Then (a )(an )(e)mail(s) has/have been sent to :to with the subject :subject:
+   * @Then (a )(an )(e)mail(s) has/have been sent with the attachment(s) :attachments:
+   * @Then (a )(an )(e)mail(s) has/have been sent to :to with the attachment(s) :attachments:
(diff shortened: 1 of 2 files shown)


AlexSkrypnyk:
Hey @chrisolof — thanks for putting this together! Attachment checking is a genuinely useful feature.

After reviewing, we've found that the [`drevops/behat-steps`](https://github.com/drevops/behat-steps) package already provides this functionality via its `EmailTrait`:

```gherkin
Then the file "document.pdf" should be attached to the email with the subject "Your document"
Then the file "report.xlsx" should be attached to the email with the subject containing "Monthly Report"
```

Rather than expanding every existing mail step with attachment parameters, we'd recommend using `drevops/behat-steps` alongside this extension — it covers attachments (and quite a bit more) with standalone, composable steps.

Closing this one in favour of that approach, but we appreciate the contribution!

PR #674, opened 2024-07-24
Use mink/webdriver-classic-driver.

saidatom:
Use [mink/webdriver-classic-driver](https://github.com/minkphp/webdriver-classic-driver) instead of [behat/mink-selenium2-driver](https://github.com/minkphp/MinkSelenium2Driver)


saidatom:
@jhedstrom align with main branch

Issue #493, opened 2018-07-09
Allow to add extra entries to the text and selectors in behat.yml


Sweetchuck:
Allow to add extra entries to the selectors in behat.yml

The [text](https://github.com/jhedstrom/drupalextension/blob/d8fcc7c0f4c251c9911bfbd6beac74ab6df2af9e/src/Drupal/DrupalExtension/ServiceContainer/DrupalExtension.php#L116) and [selectors](https://github.com/jhedstrom/drupalextension/blob/d8fcc7c0f4c251c9911bfbd6beac74ab6df2af9e/src/Drupal/DrupalExtension/ServiceContainer/DrupalExtension.php#L140) is limited to the following items:
```yml
default:
    extensions:
        Drupal\DrupalExtension:
            text:
                log_in: 'foo',
                log_out: 'foo',
                password_field: 'foo',
                username_field: 'foo',
            selectors:
                message_selector: 'foo'
                success_message_selector: 'foo'
                warning_message_selector: 'foo'
                error_message_selector: 'foo'
```
It would be very useful to add custom "text" and "selectors" entries to the _behat.yml_, and those could be used in custom step definitions.

tame4tex:
We use this patch and find it very useful. Just wondering what is preventing it from getting merged in?

jhedstrom:
Thanks!

PR #685, opened 2025-01-10
Issue #493 - 5.x Allow to add extra entries to the text and selectors

Sweetchuck:
Issue #493 - Allow to add extra entries to the text and selectors i behat.yml.


jhedstrom:
Thanks!

PR #686, opened 2025-02-21
fix(php): Fix PHP 8.4 nullable type deprecation warning

klausi:
Deprecated:  Drupal\DrupalExtension\ServiceContainer\DrupalExtension::__construct(): Implicitly marking parameter $processor as nullable is deprecated, the explicit nullable type must be used instead in /var/www/html/backend/vendor/drupal/drupal-extension/src/Drupal/DrupalExtension/ServiceContainer/DrupalExtension.php on line 41


klausi:
Uploading stable patch file for composer patches.

[behat-drupal-extension-686.diff.txt](https://github.com/user-attachments/files/18907740/behat-drupal-extension-686.diff.txt)

ptmkenny:
Also fixed here: https://github.com/jhedstrom/drupalextension/pull/682

PR #689, opened 2025-07-01
Upgrade symfony/http-client to compatibility with Symfony 7

lisotton:
In a previous PR, the deps from Symfony were updated, but http-client is forcing to stay in Symfony 6, which prevents to go to Drupal 11.


jhedstrom (approved):

jhedstrom:
Thanks!

PR #695, opened 2025-11-17
Wait for page load after login to fix race condition

juanjol:
JavaScript tests randomly fail with "Unable to determine if logged in"
  errors. This happens because `loggedIn()` is called immediately after
  clicking the login button, before the page redirect and DOM are fully
  loaded. The issue is more frequent in CI environments like Jenkins.

  ## Solution proposed

  Added "waits ups to X seconds" after login submission to ensure the page transition completes:
  - Wait up to 3 seconds for URL to change (redirect started)
  - Wait up to 3 seconds for body element to appear (page loaded)

## Summary by CodeRabbit

* **New Features**
  * Added a configurable post-login wait (configurable timeout; 0 disables waiting) to control how long the system waits for the page to update after submitting credentials.

* **Bug Fixes**
  * Improved login reliability by waiting for the URL and page content to update and for a logged-in indicator to appear before confirming authentication.


src/Drupal/DrupalExtension/Manager/DrupalAuthenticationManager.php
```diff
@@ -86,9 +86,24 @@ public function logIn(\stdClass $user)
...
         }
 
+        // Store the current URL before clicking login.
+        $loginUrl = $this->getSession()->getCurrentUrl();
+
         // Log in.
         $submit->click();
 
+        // Wait for URL change after login (max 3 seconds).
+        $timeout = microtime(true) + 3;
```
AlexSkrypnyk: this option needs to be configurable and should be set to `0` by default to allow users to opt-in.

AlexSkrypnyk (changes requested):

AlexSkrypnyk:
@coderabbitai
1. Suggest fixes for code lint issues.
2. Suggest adding a gate to enable this only for JS driver
3. Suggest tests

AlexSkrypnyk:
Hey @juanjol @EduardoMorales-mrm — thanks for this contribution! It's been a while since the last update here, so we've picked up the work and iterated on it.

We've opened #798 which builds on your changes with some improvements: moved the config to a top-level `login_wait` parameter (out of the `text` section), restored the three-phase wait with independent timeouts, added PHPUnit and Behat tests (including a Selenium test with a slow-login simulator), and fixed a fresh-build issue in `ahoy provision`. Your commits are preserved with original authorship.

We'll close this PR once the new one is merged. Appreciate the initial work!

AlexSkrypnyk:
Landed via #798. Thanks again @juanjol @EduardoMorales-mrm!

Issue #697, opened 2025-11-21
Tests are failing


rsanzante:
The tests are failing for different reasons:

- Testing two different languages confuses the parser
- 7.4 is not compatible with current code
- Some problems with D11

PR to be added.

rsanzante:
As said [here](https://github.com/jhedstrom/drupalextension/pull/699#issuecomment-3685788390), it seem,s Behat fixed he problem with the parser and the languages, so no need for that hack anymore. I'm closing #698 in favor of #699.

PR #699, opened 2025-12-21
Remove PHP 7.4 and Drupal 9 testing from CI.

AlexSkrypnyk:
Related to #697

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - php_version: "7.4"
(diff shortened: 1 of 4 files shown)


rsanzante:
So if I'm not wrong this also fixes #697, right?

It look good to me. I was surprised you didn't hit the problem when testing two different languages that I hit, but it seems Behat itself fixed that because it doesn't happen now in the commits it happens a few weeks ago.

Given that, I'm favouring this PR over mine because is cleaner (no need the obscure hack of testing English features and Spanish features separately) and covers the same fixes and more. I would merge it.

Issue #500, opened 2018-08-14
Make `ConfigContext::setConfig()` protected


pfrenssen:
As commented in #498 we should make the `ConfigContext::setConfig()` method protected. This is the helper method that actually sets the configuration. The class already has a public method for this purpose: `::setBasicConfig()` which has a step definition associated with it.

We don't need 2 public methods that do exactly the same thing.

PR #795, opened 2026-04-14
Allow scalar and special values to be set.

idimopoulos:
Ok, so we started having issues after updating to the latest version of Drupal extension, mainly due to https://github.com/jhedstrom/drupalextension/pull/779
In that MR, following the suggestion of https://github.com/jhedstrom/drupalextension/issues/500 , the `setConfig` was converted to a protected method with the argument that the `setBasicConfig` and `setComplexConfig` would suffice to properly cover the changes done by `setConfig`, thus there is no point having 2 (3) public methods doing the same thing.
However,
1. `::setBasicConfig` accepts only string $value, so booleans, integers, null etc are problematic here :/ 
2. `::setComplexConfig` accepts a TableNode and while JSON is supported, the table structure disallows to set direct values :/ so `page: false` is not really possible, because you would have something like 
```yaml
page:
  foo: bar
```
For example, I do not see how I can set my `antibot.settings` key `enabled` to FALSE. 

Proposed resolution: I would say, revert the change of visibility, but respecting previous decisions (and because it is a bit easy to be honest as we are accustomed with these steps due to our project), to have an extra step that allows explicit JSON values to be passed.
I have provided an example step (the step definition name could be better to be honest, but that is just a raw version).

## Summary by CodeRabbit

* **Tests**
  * Extended configuration testing capabilities to support multiple data types including booleans, numbers, and null values.
  * Added new test scenarios verifying typed configuration can be properly set, validated, and reset between test iterations.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/ConfigContext.php
+++ b/src/Drupal/DrupalExtension/Context/ConfigContext.php
@@ -108,6 +108,43 @@ class ConfigContext extends RawDrupalContext implements TranslatableContext {
   }
 
   /**
+   * Sets a configuration item from a JSON-encoded value.
+   *
+   * Unlike setBasicConfig() which only accepts strings, this step accepts any
+   * JSON-encodable value, making it possible to store booleans, integers,
+   * null, and arrays directly in configuration.
+   *
+   * @param string $name
+   *   The name of the configuration object.
+   * @param string $key
+   *   Identifier to store value in configuration.
+   * @param string $value
+   *   A JSON-encoded scalar or string. Must be valid JSON. Due to how behat
+   *   handles arguments, [] and {} should not be used as value here, because
+   *   they would be wrongly converted in a variable. Use ::setComplexConfig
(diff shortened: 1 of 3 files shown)


AlexSkrypnyk:
Hey @idimopoulos — thanks for flagging this and taking the time to put together a PR. You're absolutely right: this was an oversight on my end when I made `setConfig()` protected in #779. The assumption that `setBasicConfig()` and `setComplexConfig()` would cover all the cases was wrong — they clearly don't when you need to set booleans, integers, or null values.

I've taken a slightly different approach in #796. Rather than adding a new step definition, I added type coercion directly into the existing `setBasicConfig()` step — so `"true"`, `"false"`, `"null"`, and numeric strings get automatically converted to their native PHP types. The idea is that the Gherkin stays natural and there's no new step to learn. I also fixed `setComplexConfig()` while I was at it, since it had a similar issue with `null` values.

Going to close this one in favour of #796. Cheers!

PR #851, opened 2026-05-05
Allow bundle fields defined in hooks to be recogised.

idimopoulos:
This is the equivalent of https://github.com/jhedstrom/DrupalDriver/pull/356
Now, in Drupal extension v6 and driver v3, the issue resides on the extension side rather than the driver. 

Unfortunately @AlexSkrypnyk , the new versions still had the same issue as described in the other PR with both the meta entities and the URI of `rdf_sync`. And of course, as mentioned, the meta entities can be countered because they are dynamic reverse references, but the URI with the custom storage is quite important. Awaiting whether something like that suits the purpose. 

Maybe a test is needed?

## Summary by CodeRabbit

* **Bug Fixes**
  * Entity field parsing now correctly recognizes bundle-specific fields, preventing validation errors during entity creation.

* **Refactor**
  * Field validation mechanisms updated to incorporate bundle context awareness for more accurate field classification across entity types.

### DIFF
--- a/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
+++ b/src/Drupal/DrupalExtension/Context/RawDrupalContext.php
@@ -424,7 +424,7 @@ class RawDrupalContext extends RawMinkContext implements DrupalAwareInterface, D
       throw new \RuntimeException(sprintf('The active Drupal driver "%s" does not support field inspection.', $driver::class));
     }
 
-    $parser = $this->getFieldParser($stub->getEntityType(), $driver->getCore()->getFieldClassifier());
+    $parser = $this->getFieldParser($stub->getEntityType(), $driver->getCore()->getFieldClassifier(), $stub->getBundle());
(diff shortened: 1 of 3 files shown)


saidatom:
Confirmed that this works as expected. The implementation looks good to me. Thanks!

AlexSkrypnyk:
Hey @idimopoulos - thank you for this contribution! I've picked up where you left off.

I've opened #852 which builds on your work with some additional changes: bundle-aware regression tests for both parsers (covering each F6-F9 predicate), a synced F1-F9 explanation comment across `EntityFieldParser` and `LegacyEntityFieldParser`, and `null` literal casing aligned with the Drupal coding standard.

I'll close this PR once the new one lands. Your contribution is attributed in the new PR. Thanks again!

AlexSkrypnyk:
Landed via #852. Thanks again, @idimopoulos!

